PR #2, opened 2015-02-09
Audio slide

Tag0Mag0:


app/views/think_feel_do_engine/shared/_jplayer.html.erb
```diff
@@ -0,0 +1,49 @@
...
+        To play the media you will need to either update your browser to a recent version or update your <a href="http://get.adobe.com/flashplayer/" target="_blank">Flash plugin</a>.
+      </div>
+    </div>
+  </div>
+</div>
+<p><%= audio_url %></p>
+<script>
+  $("#jquery_jplayer_1").jPlayer({
+    ready: function () {
+      debugger;
```
michaelwehrley: @Tag0Mag0 is this still in there? You may consider removing it unless you need it.

PR #5, opened 2015-02-10
Add a stepped_on date column to memberships. 

gwensmu:
Addresses [#87752718] and [#87752774]. Added a migration, revised specs, added dates when participants were stepped to the clinical patient reports in the dashboard.


eschlange:
Ready to merge.

PR #8, opened 2015-02-10
assessments: delete controller, refactor specs

ericcf:
@michaelwehrley


spec/features/participant/assessments_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+        end.to change { PhqAssessment.count }.by(1)
+
+        expect(page).to have_text("Working Alliance Inventory")
+
+        expect do
+          click_on("submit")
+        end.to change { WaiAssessment.count }.by(1)
+
+        expect(page).to have_text("Thank you")
+      end
```
gwensmu: is there a case where the user has an invalid (eg expired) token? if so should there be a spec for that here?

ericcf: Tokens do not expire, but there is a spec to handle the case of one that is invalid: https://github.com/cbitstech/think_feel_do_engine/blob/master/spec/controllers/think_feel_do_engine/participants/assessments_controller_spec.rb

gwensmu: Cool. Is there any difference between the phq token and the wai one that would make the assessments_controllers_spec not also cover the wai token case?

ericcf: The only difference is the `token_type` string attribute, which is either `phq9` or `wai`, respectively. Functionally they work the same. Also, the only token string that gets passed in via request parameter is for phq. The wai tokens are looked up based on participant_id and release_date.

gwensmu: Ok thank you for explicating. Looks good to me.

ericcf:
merged: https://github.com/cbitstech/think_feel_do_engine/commit/2430646a29b2c43a04568390b08eece3c5ff3d49

PR #14, opened 2015-02-11
Remove double popup from activity builder form

gwensmu:
Remove extra call to show social sharing verify
popup to user; rely on callback in listener to
show popup for building an activity.


ericcf:
@gwensmu which Pivotal story does this refer to? Also, please try to follow the suggested commit message format in the Thoughtbot guide. Thanks!

gwensmu:
@ericcf pivotal # 87991792
roger on commit message format.

app/views/think_feel_do_engine/activities/_build_activity.html.erb
```diff
@@ -1,4 +1,4 @@
-<%= form_for current_participant.activities.build, url: create_path, method: "post", html: { role: "form", class: "activity_form", id: "activity-form", onsubmit: "validatePublic(event,'activity_shared_item_true')"} do |f| %>
+<%= form_for current_participant.activities.build, url: create_path, method: "post", html: { role: "form", class: "activity_form", id: "activity-form"} do |f| %>
```
ericcf: @gwensmu @Tag0Mag0 @eschlange does deleting the onsubmit listener entirely break any intended functionality?

Tag0Mag0: It triggers the validation alert. Is there another way that you are validating that a participant wants to make something public, @gwensmu ?

gwensmu:
@ericcf i am not sure, that is why i chose to use it over the version often called in onSubmit in other forms

ericcf:
I don't understand. What did you choose to use?

gwensmu:
https://github.com/cbitstech/think_feel_do_engine/blob/master/app/assets/javascripts/think_feel_do_engine/validate_social_sharing.js has two methods:

validatePublic is often called in onSubmit
validatePublicNoEvent is getting called by an event listener in 

https://github.com/cbitstech/think_feel_do_engine/blob/master/app/assets/javascripts/think_feel_do_engine/activities/forms.js

which is why there was a double popup. I chose to treat the forms listeners as a default behavior and the event in onSubmit as the duplicate. That choice was based on "it worked." @eschlange @Tag0Mag0 context welcome.

ericcf:
I see validatePublicNoEvent getting added in https://github.com/cbitstech/think_feel_do_engine/blob/master/app/assets/javascripts/think_feel_do_engine/activities/activity_tracker.js

Is that what you meant?

gwensmu:
That is what I meant. The form was triggering two separate validation
popups via separate methods that displayed the same text. One via the
listener in activity_tracker, one via the text I deleted in the form
template in the commit.

ericcf:
Ok, it all makes sense now! Thanks.

ericcf:
Merging in master, rspec-ing, then merging the pull...

ericcf:
Merged from the command line. Not sure why this didn't close automatically.

PR #16, opened 2015-02-12
[Finished #88007666] Updated the patient dashboard show page to use the ...

eschlange:
...membership relative to the current group view. Updated specs, added a helper method for use on the patient show page.

### DIFF
--- a/app/views/think_feel_do_engine/coach/patient_dashboards/_patient.html.erb
+++ b/app/views/think_feel_do_engine/coach/patient_dashboards/_patient.html.erb
@@ -1,7 +1,7 @@
 <% if can? :show, patient %>
   <tr id="patient-<%= patient.id %>"
-      class="<%= 'inactive' if patient.active_membership &&
-              patient.active_membership.end_date < Date.current %>
+      class="<%= 'inactive' if view_membership(patient, @group) &&
+              view_membership(patient, @group).end_date < Date.current %>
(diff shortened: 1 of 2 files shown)


app/views/think_feel_do_engine/coach/patient_dashboards/_patient.html.erb
```diff
@@ -1,7 +1,7 @@
 <% if can? :show, patient %>
   <tr id="patient-<%= patient.id %>"
-      class="<%= 'inactive' if patient.active_membership &&
-              patient.active_membership.end_date < Date.current %>
+      class="<%= 'inactive' if view_membership(patient, @group) &&
```
ericcf: What do you think about calling `patient.memberships.find_by(group: @group)` instead of `view_membership(patient, @group)` or alternatively passing in a variable from the controller for this membership?

eschlange: @ericcf  The patient dashboard page displays an array of participants, consolidating their memberships into a data structure, then pulling it out of the data structure seems unnecessarily complex. With regards to using patient.memberships.find_by(group: @group) instead of view_membership, I toyed with that idea at first, but it seemed more comprehensive to use a more descriptive method that we can use across views. It specifically ties together a patient and group within the parameters.

ericcf: Ok, thanks! :shipit:

PR #19, opened 2015-02-12
Ems 85239860 tfdengine travis spec fix

cbrenner04:
one test still fails intermittently because of the footer overlapping the button. trying to come up with a way to fix nothing working yet. Otherwise all tests are passing.


cbrenner04:
this is being closed due to rebase and merge errors

PR #21, opened 2015-02-13
Implement view membership helper in patient dahsboard

Tag0Mag0:
* Update inactive status view logic for relative membership

[Finishes #88267970]


ericcf:
@Tag0Mag0 do any specs need updating, or is this good to go?

Tag0Mag0:
@ericcf There are multiple specs that check the status and membership dates that didn't need to be updated. I suppose we could check whether the helper is returning that specific membership rather than some other. What do you think?

ericcf:
I think this is reasonable. We need a review of our specs anyway, so let's handle it then.

ericcf:
Will update the pull request in a couple of minutes.

ericcf:
Merged. Closing.

PR #22, opened 2015-02-13
Cjb 85239860 new travis spec fix

cbrenner04:


cbrenner04:
I think this is ready because the last commit from the old branch passed travis? I think that is why this isn't on travis? I cannot be sure.

ericcf:
alright...merging

PR #24, opened 2015-02-13
List of lessons only shows current arm options

gwensmu:
* edit controller for coach messages to add group var
* now list of lessons will only show options for the
  current group of participants

[#84857732]


app/controllers/think_feel_do_engine/coach/messages_controller.rb
```diff
@@ -20,6 +20,7 @@ def index
       def new
         authorize! :new, Message
 
+        @group = @group = Group.find(params[:group_id])
```
ericcf: @gwensmu two questions: 1. is there a reason for the double assignment here? 2. why is this assignment necessary when the `before_action` in the controller appears to already handle it? Thanks.

gwensmu: @ericcf double assignment is a typo. On reflection though I'm going to
close this pull request, not sure that the seed fixtures currently have
enough in them to replicate the bug being seen on staging. Going to verify
then re-open.

PR #29, opened 2015-02-16
Jah participant extension

Tag0Mag0:
Adding this to Participant model to facilitate conditional coach assignment logic in Dashboard.


ericcf:
@Tag0Mag0 mind if I delete?

Tag0Mag0:
Too late!

PR #30, opened 2015-02-16
Update Thought Views

michaelwehrley:
* Update data-item-type of
  span tags from 'activity' to
  'thought' in order to fix
  failing spec in think_feel_do_so
* [Updates: #86635422]


eschlange:
looks good.

PR #33, opened 2015-02-16
handle Provider lacking data class with alert

ericcf:
* show user-friendly message rather than 500-style error

[#88459130]


ericcf:
@michaelwehrley thanks, I like that suggestion.

PR #35, opened 2015-02-17
Msw cancel button

michaelwehrley:
Change Cancel HREF

* Cancel button on the "New" page  
  references a dynamic route  
  depending if the Content Provider  
  is being created independently  
  or within a Content Module
* [Finishes: #88450180]


ericcf:
Yay

PR #37, opened 2015-02-17
Remove Effect From Thoughts Table

michaelwehrley:
* Remove "Effect" column.
* Harmful thoughts table only displays  
  harmful thoughts - the participant  
  can no longer edit the "effect" to  
  'helfpul' or 'neither' from this page.
* [Finishes: #88524260]


eschlange:
Looks good.

PR #38, opened 2015-02-18
configure with Brakeman, ignoring false positives

ericcf:
* several `link_to` expressions are flagged incorrectly by Brakeman


cbrenner04:
can someone review and merge please?

think_feel_do_engine.gemspec
```diff
@@ -37,4 +37,5 @@ Gem::Specification.new do |s|
   s.add_development_dependency "poltergeist", "~> 1.5"
   s.add_development_dependency "database_cleaner", "~> 1.3"
   s.add_development_dependency "jasmine-rails", "0.10.0"
+  s.add_development_dependency "brakeman"
```
eschlange: Should this Gem version be unbound?

eschlange:
:+1:

think_feel_do_engine.gemspec
ericcf: That's intentional, because I always want the latest version available. Because it's checking for vulnerabilities, it's most valuable to us if we can keep installing improved releases.

PR #40, opened 2015-02-18
Change available links in message form

gwensmu:
* Coach may select modules from the set
  of tasks currently assigned to the group
  to include in a message to a participant

[#84857732]


michaelwehrley:
@gwensmu Just checking in on this.  Did you decided to an/or were you able to create any specs?  Thanks.

gwensmu:
@michaelwehrley working on another story right now, but -- the specs that are there are a bit fudged, in that they don't test an actual link, they test placeholder text in the message. i'd want to amend the current specs for the form to test the link generation but that's a bigger task testing more than just this change. thoughts?

michaelwehrley:
@gwensmu If I remember correctly, I thought you made a good point about stopping testing scope creep and instead just write a test that would test that we only see ```titles``` that are now part of assigned tasks and a check to make sure that ```titles``` that are not assigned are not visible.

gwensmu:
@michaelwehrley sure i'll just write a spec for the helper then.

michaelwehrley:
@gwensmu Thanks, I think that would be helpful.

gwensmu:
@michaelwehrley updated specs, @ericcf sorry for not following commit message format, was an accident

ericcf:
@gwensmu no worries. @michaelwehrley I'll let you make the call on merging this.

michaelwehrley:
@gwensmu Instead of me being Mr. Bossy Pants, I figured, I would try to help out.  You can take a look at the commit: https://github.com/cbitstech/think_feel_do_engine/commit/b6aaddcbeb56300a460d5d8580d8b7d9470e0211 and the merge.  Thanks!

PR #43, opened 2015-02-19
Update Completion Calculation

michaelwehrley:
* Display only 'is_scheduled' activities in the  
  Daily Summary and the completion calculation.
* Add model spec to .scheduled scope.
* Update feature spec to display new calculation.
* [Finishes: #88471620]


eschlange:
:+1:

PR #48, opened 2015-02-19
Check Validity of .unfinished_awake_periods

michaelwehrley:
* Create test to demonstrate that the scope  
  .unfinished_awake_periods is actually  
  scoping to the correct content.
* After some consternation, an arel solution,  
  while outer joins are possible, was  
  not found and/or written.

[Finishes: #88451248]

### DIFF
--- a/spec/models/awake_period_spec.rb
+++ b/spec/models/awake_period_spec.rb
@@ -70,13 +70,13 @@ describe AwakePeriod do
         participant.awake_periods.create(
           start_time: datetime,
           end_time: datetime.advance(hours: 1))
-      end.to change{ participant.unfinished_awake_periods.count }.by(1)
+      end.to change { participant.unfinished_awake_periods.count }.by(1)
 
       expect do
         participant.activities.create(
           activity_type: activity_types(:jogging),
           start_time: datetime)
-      end.to change{ participant.unfinished_awake_periods.count }.by(-1)
+      end.to change { participant.unfinished_awake_periods.count }.by(-1)
     end
   end
 end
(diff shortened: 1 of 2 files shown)


spec/models/awake_period_spec.rb
```diff
@@ -59,5 +59,23 @@
 
       expect(scoped_awake_periods.count).to eq 2
     end
+
+    it ".unfinished_awake_periods returns awake periods for which there are no activities with corresponding start_time" do
+      unfinished_awake_periods_count = participant3.unfinished_awake_periods.count
```
ericcf: @michaelwehrley what do you think about using the `change` matcher? https://www.relishapp.com/rspec/rspec-expectations/v/3-2/docs/built-in-matchers/change-matcher

spec/models/awake_period_spec.rb
```diff
@@ -1,7 +1,7 @@
 require "rails_helper"
 
 describe AwakePeriod do
-  fixtures :participants, :awake_periods
+  fixtures :participants, :awake_periods, :activity_types
 
   describe "validations" do
```
ericcf: @michaelwehrley a more conventional naming scheme here would be
```
describe "#unfinished_awake_periods" do
...
```
and then
```
it "returns awake periods for which there are no activities with corresponding start_time"
```
`#` denotes an instance method and `.` denotes a class method.

spec/models/awake_period_spec.rb
michaelwehrley: sounds good to me.

spec/models/awake_period_spec.rb
michaelwehrley: ahh yes, this one got me b/c it is a scope on an instance.  Good catch.

michaelwehrley:
@ericcf Thanks! :ice_cream:

PR #51, opened 2015-02-20
Refactor patient dashboard views

gwensmu:
* Separate partials into two views
  one for active participants, and one
  for inactive, to streamline logic
* Update specs to contend with errors caused
  by javascript modal rendering

[#88235970]


eschlange:
Looks good, much more comprehensive than before.

PR #52, opened 2015-02-23
Add logic to check for no completed activities

gwensmu:
* Completion score only calculates a value if
  scheduled activities have been complted.
* Otherwise simply displays 0
* Fixes a division by zero error.

[#88803632]

### DIFF
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -155,6 +155,10 @@ class Activity < ActiveRecord::Base
     end
   end
 
+  def self.completion_score
+    completed.count > 0 ? (completed.count * 100 / count).round : 0
+  end
+
   private
 
   def actual_accomplishable_updates
--- a/app/views/think_feel_do_engine/activities/_daily_summary.html.erb
+++ b/app/views/think_feel_do_engine/activities/_daily_summary.html.erb
@@ -26,7 +26,11 @@
         </p>
         <p>
           Completion Score:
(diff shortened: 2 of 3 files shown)


ericcf:
@gwensmu is there a spec that fails when the completed activities count == 0?

gwensmu:
@ericcf no, i don't know how i would write a spec that fails since the view logic now catches that case.

gwensmu:
@ericcf do you mean, a spec that passes even when completed activities count == 0?

gwensmu:
@eschlange just made the related point in an unrelated convo that this kind of logic shouldn't be in the view - i can do the calculation in the model, and just reference it in the view -- dunno if that refactoring makes sense in the context of this bug fix @ericcf

ericcf:
@gwensmu specs should be written such that they fail before the change has been made, then pass once the change has been made.

WRT where the logic lives, ideally it should be in a Presenter-type class or in the Model class.

Thanks.

gwensmu:
@ericcf, right, i am asking a process/priorities question. should i do this
refactoring now, as part of the time i am allocating to this bug fix.

ericcf:
Go ahead and do the refactor now while the code is fresh in your mind!

gwensmu:
@ericcf added a second commit with the changes, thanks for your input!

eschlange:
Looks good.

PR #57, opened 2015-02-24
import remaining shared reports

ericcf:
[#88546774]

### DIFF
--- a/spec/models/ability_spec.rb
+++ b/spec/models/ability_spec.rb
@@ -44,14 +44,17 @@ describe Ability do
       it "can view all Tasks" do
         expect(researcher.can?(:read, Task)).to eq true
       end
+
+      it "can read Reports" do
+        expect(researcher.can?(:read, "Reports")).to eq true
+      end
     end
 
     describe "Non-superuser, multi-role Users" do
       it "share all Roles' abilities" do
         expect(multi_role.can?(:manage, BitCore::ContentModule)).to eq true
         expect(multi_role.can?(:manage, Task)).to eq true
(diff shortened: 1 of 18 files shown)


ericcf:
Related to: https://github.com/cbitstech/think_feel_do_dashboard/pull/10

spec/models/ability_spec.rb
```diff
@@ -44,14 +44,17 @@ def role(name)
...
+      end
     end
 
     describe "Non-superuser, multi-role Users" do
       it "share all Roles' abilities" do
         expect(multi_role.can?(:manage, BitCore::ContentModule)).to eq true
         expect(multi_role.can?(:manage, Task)).to eq true
-        expect(multi_role.can?(:manage, ThinkFeelDoDashboard::Reports::Comment))
-          .to eq true
+        expect(multi_role.can?(:read, "Reports")).to eq true
```
gwensmu: Does it make sense for authorization specs to test if a user who shouldn't be able to read reports can't? I don't see that pattern in the overall spec, is that covered elsewhere?

ericcf: Yes it does. No I don't think it's covered.


--
Eric Carty-Fickes


-------- Original message --------

gwensmu:
Cool, I'd like to see it here at least for this spec.

ericcf:
@michaelwehrley yes: https://github.com/cbitstech/think_feel_do_dashboard/pull/10/files#diff-df86902ce9dceb01164aa1b2f4e97aed

PR #63, opened 2015-02-27
Msw is reviewed

michaelwehrley:
* Create migration to add ```is_reviewed``` boolean  
  to activities table.
* Add validation so both predicted value attributes  
  must exist if one exists.
* Update activity model specs to add  
  default values to pass validations.  
  Merging overrides default values.
* Write methods that describe the state of an activity.
* Create ```monitored?``` with specs.
* Create ```planned?``` with specs.
* Create ```reviewed_and_complete?``` with specs.
* Create ```reviewed_and_incomplete?``` with specs.
* Remove is_monitored from activities table.
* Remove references of database column  
  ```is_monitored``` from methods.
* Update views with new monitored? method.

### DIFF
--- a/spec/models/activity_spec.rb
+++ b/spec/models/activity_spec.rb
@@ -9,7 +9,11 @@ RSpec.describe Activity do
     def sleeping(attributes = {})
       Activity.create!({
         participant: participant,
-        activity_type: activity_types(:sleeping)
+        activity_type: activity_types(:sleeping),
+        predicted_accomplishment_intensity: 5,
+        predicted_pleasure_intensity: 5,
+        actual_accomplishment_intensity: 5,
+        actual_pleasure_intensity: 5
       }.merge(attributes))
     end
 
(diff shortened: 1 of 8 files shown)


spec/models/activity_spec.rb
```diff
@@ -150,6 +154,146 @@ def sleeping(attributes = {})
...
+        ).planned?).to be false
+      end
+
+      it "returns false if predictions don't exist" do
+        expect(running(
+          predicted_accomplishment_intensity: nil,
+          predicted_pleasure_intensity: nil
+        ).planned?).to be false
+      end
+
```
cbrenner04: @michaelwehrley do we want a "returns false if monitored is true"?

cbrenner04:
@michaelwehrley this looks great. Just one comment about possibly needing an additional spec. It may be redundant though so just let me know. I will merge once we discuss that. Nice job!

PR #73, opened 2015-03-02
Make lesson pages more printable

gwensmu:
* Create custom printable template for lessons
  with pared-down formatting
* In printable template, replace multimedia with
  a text placeholder
* Add inline attribute to all divs in print.css to
  prevent content from getting cut off
* Override bootstrap container for Firefox

[#88866314]


ericcf:
Seems good to me. :green_apple:

PR #74, opened 2015-03-02
Ecf reports

ericcf:


michaelwehrley:
I didn't know about the ```take_while``` and ```drop_while``` methods.  Nice.  It looks like some intense refactoring.  :+1: :imp:

PR #76, opened 2015-03-03
Ecf 82515802 event exports

ericcf:


app/models/think_feel_do_engine/reports/reporter.rb
```diff
@@ -0,0 +1,24 @@
+require "csv"
+
+module ThinkFeelDoEngine
+  module Reports
+    # Produces formatted versions of data produced by a data collector.
+    # The collector is expected to have two methods:
+    # `columns` returns a collection of column names
+    # `all` returns a collection of Hashes whose keys match the column names
+    class Reporter
```
gwensmu: nice DRY in action

spec/models/think_feel_do_engine/reports/reporter_spec.rb
```diff
@@ -0,0 +1,18 @@
+require "spec_helper"
+require_relative "../../../../app/models/think_feel_do_engine/reports/reporter"
+
+module ThinkFeelDoEngine
```
gwensmu: hard not to feel good about this. almost want it to be a cbitz gem or something, seems like it'd be super useful to have across probably any rails project we build here

gwensmu:
🏫🌵🏫 looks great

PR #77, opened 2015-03-03
Fix Error For Activities Helper

michaelwehrley:
Fix Error For Activities Helper

* Update helper to check if activities that were planned is zero.
* Refactor helper text code.

[#89418002] [#89044666]

### DIFF
--- a/app/helpers/think_feel_do_engine/activities_helper.rb
+++ b/app/helpers/think_feel_do_engine/activities_helper.rb
@@ -21,6 +21,8 @@ module ThinkFeelDoEngine
       "(#{help_text(activities.were_planned)})"
     end
 
+    private
+
     def help_text(activities)
       if activities.were_planned.count > 0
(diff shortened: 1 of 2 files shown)


app/helpers/think_feel_do_engine/activities_helper.rb
```diff
@@ -17,22 +17,27 @@ def average_intensity_difference(activities, attribute)
...
+      if activities.were_planned.count > 0
+        "You completed #{activities.reviewed_and_complete.count} out of "\
+        "#{scheduled_message(activities.were_planned)} that you scheduled."
       else
-        "Completion Score: Not Available (No activities were scheduled.)"
+        "No activities were scheduled."
       end
     end
 
     def percent_complete(activities)
```
ericcf: @michaelwehrley should this or any of the other helper methods be private? Are they used elsewhere?

michaelwehrley: good point.

PR #80, opened 2015-03-03
Ems 89311948 Group Dashboard - Lesson Panel

eschlange:
* Added lesson breakdown for a group scope.
* Included a total number of lessons read.
* Included list of participants that completed the lesson.
* Included list of participants whose lessons were not yet complete.
* Added specs for all of the new helper methods and model utils.

[Finished #89311948]

### DIFF
--- a/app/helpers/think_feel_do_engine/coach/group_dashboard_helper.rb
+++ b/app/helpers/think_feel_do_engine/coach/group_dashboard_helper.rb
@@ -58,36 +58,43 @@ module ThinkFeelDoEngine
         end
       end
 
+      # Disabling next style, false positive.
+      # rubocop:disable Style/Next
       def display_lesson_details_by_week(group, week_number)
         group.learning_tasks.collect do |task|
           if week_number == week_of_task(task)
-            task.bit_core_content_module.content_providers.collect do |content_provider|
+            task.bit_core_content_module.content_providers
+              .collect do |content_provider|
               "<tr><td>#{content_provider.source_content.title}</td>"\
                 "#{participants_that_read_lesson(task)}</tr>"
             end.join
           end
         end.join
       end
+      # rubocop:enable Style/Next
 
       def participants_that_read_lesson(task)
         total_assigned = 0
         total_read = 0
         task.task_statuses.each do |status|
           total_assigned += 1
-          total_read += 1 if !status.completed_at.nil?
+          total_read += 1 unless status.completed_at.nil?
         end
         "<td>#{total_read} of #{total_assigned} COMPLETE</td>"\
         "<td>"\
         "#{list_participant_names(task.group, task.complete_participant_list)}"\
         "</td><td>"\
-        "#{list_participant_names(task.group, task.incomplete_participant_list)}</td>"
+        "#{list_participant_names(task.group,
+                                  task.incomplete_participant_list)}</td>"
       end
 
       def list_participant_names(group, participants)
         participant_list = "<ul>"
         participants.each do |participant|
           participant_list +=
-          "<li>#{link_to participant.display_name, coach_group_patient_dashboard_path(group, participant)}</li>" if participant
+          "<li>#{link_to participant.display_name,
+                         coach_group_patient_dashboard_path(
+                           group, participant)}</li>" if participant
         end
         participant_list += "</ul>"
         participant_list
--- a/app/helpers/think_feel_do_engine/coach/group_dashboard_helper.rb
+++ b/app/helpers/think_feel_do_engine/coach/group_dashboard_helper.rb
@@ -58,6 +58,20 @@ module ThinkFeelDoEngine
         end
       end
 
+      def display_lesson_details_by_week(group)
+        content = ""
+        group.learning_tasks.each do |task|
+          task.bit_core_content_module.content_providers.collect do |content_provider|
+            puts "TEST123"
+            content_tag :h3 do
+              "#{content_provider.source_content.title}"\
+                "#{content_tag(:small, "0 of 12 complete")}"
+            end
+          end.join.inspect
+        end
+        content
+      end
+
       def comment_shared_item_description(comment)
         case comment.item.item_type
         when "Activity"
--- a/app/helpers/think_feel_do_engine/coach/group_dashboard_helper.rb
+++ b/app/helpers/think_feel_do_engine/coach/group_dashboard_helper.rb
@@ -58,18 +58,39 @@ module ThinkFeelDoEngine
         end
       end
 
-      def display_lesson_details_by_week(group)
-        content = ""
-        group.learning_tasks.each do |task|
-          task.bit_core_content_module.content_providers.collect do |content_provider|
-            puts "TEST123"
-            content_tag :h3 do
-              "#{content_provider.source_content.title}"\
-                "#{content_tag(:small, "0 of 12 complete")}"
-            end
-          end.join.inspect
+      def display_lesson_details_by_week(group, week_number)
+        group.learning_tasks.collect do |task|
+          if week_number == week_of_task(task)
+            task.bit_core_content_module.content_providers.collect do |content_provider|
+              "<tr><td>#{content_provider.source_content.title}</td>"\
+                "#{participants_that_read_lesson(task)}</tr>"
+            end.join
+          end
+        end.join
+      end
+
+      def participants_that_read_lesson(task)
+        total_assigned = 0
+        total_read = 0
+        task.task_statuses.each do |status|
+          total_assigned += 1
+          total_read += 1 if !status.completed_at.nil?
         end
-        content
+        "<td>#{total_read} of #{total_assigned} COMPLETE</td>"\
+        "<td>"\
+        "#{list_participant_names(task.group, task.complete_participant_list)}"\
+        "</td><td>"\
+        "#{list_participant_names(task.group, task.incomplete_participant_list)}</td>"
+      end
+
+      def list_participant_names(group, participants)
+        participant_list = "<ul>"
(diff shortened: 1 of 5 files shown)


app/helpers/think_feel_do_engine/coach/group_dashboard_helper.rb
```diff
@@ -58,6 +58,48 @@ def comment_item_description(comment)
         end
       end
 
+      # Disabling next style, false positive.
+      # rubocop:disable Style/Next
+      def display_lesson_details_by_week(group, week_number)
```
ericcf: What do you think about rewriting this using a partial? Something like
```
<%= render partial: "weekly_lesson_details", collection: group.learning_tasks.on_week(week_number) %>
...
<tr><td><%= content_provider.source_content.title %> <%= participants_that_read_lesson(task) %></td></tr>
```

app/helpers/think_feel_do_engine/coach/group_dashboard_helper.rb
```diff
@@ -58,6 +58,48 @@ def comment_item_description(comment)
...
+        group.learning_tasks.collect do |task|
+          if week_number == week_of_task(task)
+            task.bit_core_content_module.content_providers
+              .collect do |content_provider|
+              "<tr><td>#{content_provider.source_content.title}</td>"\
+                "#{participants_that_read_lesson(task)}</tr>"
```
ericcf: This doesn't look like valid HTML.

app/helpers/think_feel_do_engine/coach/group_dashboard_helper.rb
```diff
@@ -58,6 +58,48 @@ def comment_item_description(comment)
...
+            end.join
+          end
+        end.join
+      end
+      # rubocop:enable Style/Next
+
+      def participants_that_read_lesson(task)
```
ericcf: What about pulling out the `total_assigned` and `total_read` logic into model methods? They'll be more easily reusable and testable that way.

app/helpers/think_feel_do_engine/coach/group_dashboard_helper.rb
```diff
@@ -58,6 +58,48 @@ def comment_item_description(comment)
...
+        end
+        "<td>#{total_read} of #{total_assigned} COMPLETE</td>"\
+        "<td>"\
+        "#{list_participant_names(task.group, task.complete_participant_list)}"\
+        "</td><td>"\
+        "#{list_participant_names(task.group,
+                                  task.incomplete_participant_list)}</td>"
+      end
+
+      def list_participant_names(group, participants)
```
ericcf: What about rewriting this as a partial?

app/helpers/think_feel_do_engine/coach/group_dashboard_helper.rb
```diff
@@ -77,6 +119,16 @@ def comment_shared_item_description(comment)
           " Item Type: #{comment.item_type}"
         end
       end
+
+      private
+
+      def week_of_task(task)
```
ericcf: What do you think about making this a method on Task?

spec/models/task_spec.rb
```diff
@@ -88,4 +88,25 @@
...
+    let(:group) { groups(:group1) }
+    let(:module1) { bit_core_content_modules(:do_awareness) }
+    let(:learning_module) do
+      bit_core_content_modules(:slideshow_content_module_1)
+    end
+    let(:participant1) { participants(:participant1) }
+
+    it "incomplete_participant_list should return a list of participants "\
+    "that have not completed this task " do
+      expect(task1.incomplete_participant_list.count).to be >= 2
```
ericcf: Why did you write this with a `>=` comparator?

spec/helpers/think_feel_do_engine/group_dashboard_helper_spec.rb
```diff
@@ -43,6 +43,34 @@ module Coach
...
+          expect(
+            helper.display_lesson_details_by_week(group, week_number)
+          ).to eq("<tr><td>Do - Awareness Introduction</td><td>0 of 2 COMPLETE</td><td><ul></ul></td><td><ul><li><a href=\"some link\">Aqua</a></li><li><a href=\"some link\">Water</a></li></ul></td></tr><tr><td>some title</td><td>0 of 2 COMPLETE</td><td><ul></ul></td><td><ul><li><a href=\"some link\">Aqua</a></li><li><a href=\"some link\">Water</a></li></ul></td></tr><tr><td>some title</td><td>0 of 2 COMPLETE</td><td><ul></ul></td><td><ul><li><a href=\"some link\">Aqua</a></li><li><a href=\"some link\">Water</a></li></ul></td></tr><tr><td>some title</td><td>0 of 2 COMPLETE</td><td><ul></ul></td><td><ul><li><a href=\"some link\">Aqua</a></li><li><a href=\"some link\">Water</a></li></ul></td></tr><tr><td>some title</td><td>0 of 2 COMPLETE</td><td><ul></ul></td><td><ul><li><a href=\"some link\">Aqua</a></li><li><a href=\"some link\">Water</a></li></ul></td></tr><tr><td>some title</td><td>0 of 2 COMPLETE</td><td><ul></ul></td><td><ul><li><a href=\"some link\">Aqua</a></li><li><a href=\"some link\">Water</a></li></ul></td></tr>")
```
ericcf: Do you think this spec will be brittle?

app/helpers/think_feel_do_engine/coach/group_dashboard_helper.rb
eschlange: Line 69 (specifically the method participants_that_read_lesson) adds another set of cells to the table row,  resulting in valid HTML.

app/helpers/think_feel_do_engine/coach/group_dashboard_helper.rb
eschlange: I like this idea, I'll start work on this now.

app/helpers/think_feel_do_engine/coach/group_dashboard_helper.rb
eschlange: I already abstracted the lessons page into a partial, is it good practice to add another layer of partials in this case?

app/helpers/think_feel_do_engine/coach/group_dashboard_helper.rb
ericcf: Aha, makes sense.

app/helpers/think_feel_do_engine/coach/group_dashboard_helper.rb
ericcf: I think the only reason not to would be if it were costing us a big performance penalty.

eschlange: OK, cool. I'll get working on this new partial now.

eschlange: This is a difficult task, I've already spent about an hour on it. I imagine sinking another hour or two on this at least, plus time to write specs. Translating the helper logic for generating these rows is complex.

spec/models/task_spec.rb
```diff
@@ -88,4 +88,25 @@
...
+
+  describe "participant related utility methods" do
+    fixtures(:all)
+
+    let(:group) { groups(:group1) }
+    let(:module1) { bit_core_content_modules(:do_awareness) }
+    let(:learning_module) do
+      bit_core_content_modules(:slideshow_content_module_1)
+    end
+    let(:participant1) { participants(:participant1) }
```
ericcf: Where are these `let` assignments being used in this example?

app/helpers/think_feel_do_engine/coach/group_dashboard_helper.rb
eschlange: Maybe we can sit down tomorrow and discuss how this can be accomplished given this code as well as the previous partial transfer request. There is a significant amount of complex logic and I'm not sure how to translate it into partials.

app/helpers/think_feel_do_engine/coach/group_dashboard_helper.rb
eschlange: This is specific to social arms, since a task can only be set to a specific week when all participants have the same start date. This is not a valid method for non-social arms. Should this be placed into  Task in this circumstance?

app/helpers/think_feel_do_engine/coach/group_dashboard_helper.rb
ericcf: Sure. Maybe we could start as a group refactoring meeting again to get more than just my jaded opinion.

spec/models/task_spec.rb
eschlange: I found that if I used the equal matcher, if others changed fixtures then it would cause breakage.

spec/models/task_spec.rb
eschlange: I'll remove these. I used them at first, then forgot to remove them before the commit.

app/helpers/think_feel_do_engine/coach/group_dashboard_helper.rb
eschlange: Cool~

spec/helpers/think_feel_do_engine/group_dashboard_helper_spec.rb
eschlange: I agree, this was more of a placeholder until I understood the full contents of the output. I'll refactor this to check for the specific lesson, titles and links that should result from this method.

PR #84, opened 2015-03-04
Refactoring lesson panel code.

eschlange:
* Moved some helper logic into task model.
* Added task spec for new methods.
* Refactored specs to make then less brittle.

[#89311948]

### DIFF
--- a/app/models/task.rb
+++ b/app/models/task.rb
@@ -52,11 +52,7 @@ class Task < ActiveRecord::Base
 
   # Returns the count of the number of times this task was completed.
   def total_read
-    total_read = 0
-    task_statuses.each do |status|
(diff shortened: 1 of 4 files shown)


app/models/task.rb
```diff
@@ -45,6 +45,20 @@ def complete_participant_list
     end
   end
 
+  # Returns a count of the number of times this task was assigned.
+  def total_assigned
+    task_statuses.count
+  end
+
+  # Returns the count of the number of times this task was completed.
+  def total_read
```
ericcf: What do you think about refactoring this to
```
task_statuses.where.not(completed_at: nil).count
```
?

eschlange: Much cleaner! I'll change this now.

PR #85, opened 2015-03-04
Msw edit activities

michaelwehrley:
Automatically Review Activities & Update Editability

* When a participant updates an activity from  
  the 'Your Activities' page, the activity is  
  automatically reviewed.
* Model test for ```update_as_reviewed```.
* Controller spec for success and errors.
* On the 'Your Activities' page, only activities  
  that have not been  reviewed and  
  incomplete can be edited.
* Update model spec.

[Finishes: #88824148]

### DIFF
--- a/spec/controllers/think_feel_do_engine/participants/activities_controller_spec.rb
+++ b/spec/controllers/think_feel_do_engine/participants/activities_controller_spec.rb
@@ -1,0 +1,58 @@
+require "rails_helper"
+
+module ThinkFeelDoEngine
+  module Participants
+    RSpec.describe ActivitiesController, type: :controller do
+      describe "PUT update" do
+        let(:activity) { double("activity") }
+        let(:participant) { double("participant") }
+
+        context "activity updates with update_as_reviewed" do
+          before do
+            sign_in_participant participant
+            allow(participant).to receive(:activities) { Activity }
+            allow(Activity).to receive(:find) { activity }
+            allow(activity).to receive(:update_as_reviewed) { true }
+            put :update,
+                id: 1,
+                activity: {
+                  actual_accomplishment_intensity: 1,
+                  actual_pleasure_intensity: 4
+                },
+                use_route: :think_feel_do_engine,
+                format: :js
+          end
+
+          it "responds with js and visits the current page with success notice" do
+            expect(flash[:notice]).to be_present
+            expect(response.body).to eq "Turbolinks.visit(window.location);"
+          end
+        end
+
+        context "activity does not update with update_as_reviewed" do
+          before do
+            sign_in_participant participant
+            allow(participant).to receive(:activities) { Activity }
+            allow(Activity).to receive(:find) { activity }
+            allow(activity).to receive(:update_as_reviewed) { false }
+            allow(activity)
+              .to receive_message_chain(:errors, :full_messages) { ["Errors"] }
+            put :update,
+                id: 1,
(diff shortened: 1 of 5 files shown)


spec/controllers/think_feel_do_engine/participants/activities_controller_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+      describe "PUT update" do
+        let(:activity) { double("activity") }
+        let(:participant) { double("participant") }
+
+        context "activity updates with update_as_reviewed" do
+          before do
+            sign_in_participant participant
+            allow(participant).to receive(:activities) { Activity }
+            allow(Activity).to receive(:find) { activity }
+            allow(activity).to receive(:update_as_reviewed) { true }
```
gwensmu: maybe allow this mock to receive the error chain as well, and then expect not to see errors

gwensmu: what are the conditions that determine if update_as_reviwed is true or false? how are they captured in this test?

spec/models/activity_spec.rb
```diff
@@ -187,6 +187,24 @@ def sleeping(attributes = {})
...
   describe "instance methods" do
+    describe "#update_as_reviewed" do
+      let(:running) do
+        Activity.new(
+          participant: participants(:participant1),
+          activity_type: activity_types(:jogging)
+        )
+      end
+
+      it "updates activity with is_reviewed to be true" do
```
gwensmu: does anything need to happen to the activity for it to be marked is reviewed? eg, does the data change? or is it just the participant viewing the activity passively that counts as reviewing

spec/controllers/think_feel_do_engine/participants/activities_controller_spec.rb
michaelwehrley: @gwensmu Does it work that I am checking that the notice is visible: #L27?

@gwensmu The conditions for update failing are the same as ```update``` in rails in general.  I did think about placing the model code: app/models/activity.rb#L123 straight in the controller, but I couldn't figure out a way to test it unless I checked the model object in the feature test which didn't seem right.  But I am open for other options.

spec/models/activity_spec.rb
michaelwehrley: @gwensmu From my understanding, as long as they hit this route, that is, they are adding actual accomplishment and pleasure ratings.

gwensmu: @michaelwehrley cool makes sense

spec/controllers/think_feel_do_engine/participants/activities_controller_spec.rb
gwensmu: @michaelwehrley it sounds like from below that reviewed should only be true if actual accomplishment and actual pleasure are not nil, but I also see a "reviewed and incomplete" condition -- I guess what I would like to see from this spec is a way to check that "update as reviewed" won't be successful unless the correct business logic is in place.

michaelwehrley: @gwensmu Good point. :tangerine:

michaelwehrley:
@gwensmu are are these updates?  Thanks!

gwensmu:
@michaelwehrley looks pretty complete to me! 🍢

PR #88, opened 2015-03-04
fix inaccurately reported Module Sessions

ericcf:
[#88479844]


michaelwehrley:
Ahhh!  That is what I was looking for!  Thanks!

PR #89, opened 2015-03-04
Fix task status bug in toolbar

gwensmu:
* Add task-status data attributes and css to
  tool bar nav item links
* Add method to module_nav_item to obtain
  correct task status id for data attribute
* Add spec to test ModuleNavItem.task_status method
* Amend other related specs to pass

[#89512778]


eschlange:
This is some mighty-fine code.

PR #90, opened 2015-03-05
Update group dash lesson view.

eschlange:
* Refactored complex helper logic into simple partials.
* Updated specs to be less brittle.
* Removed unused helper methods.
* Reworked lesson view to have collapsable patient lists.

[Finished #89311948]


ericcf:
Thanks, this looks way better to me! FYI, there are also Rails conventions for rendering collections with partials but you can decide if you think they're cleaner: http://guides.rubyonrails.org/layouts_and_rendering.html#rendering-collections

eschlange:
@ericcf I updated the partial to use the rails convention for iteration of a collection. Please take a look and let me know if this is acceptable.

PR #94, opened 2015-03-06
Update thoughts tables.

eschlange:
* Updated thoughts tables by removing effects.
* Added three new columns to group dash thought table.
* Updated specs not to expect effects coloumn.

[Finished #89820882]

### DIFF
--- a/app/views/think_feel_do_engine/coach/group_dashboard/_thoughts.html.erb
+++ b/app/views/think_feel_do_engine/coach/group_dashboard/_thoughts.html.erb
@@ -33,9 +33,9 @@
                 </td>
                 <td><%= membership.participant.display_name %></td>
                 <td><%= thought.content %></td>
-                <td><%= (thought.pattern_title && thought.pattern_title.length > 0) ? thought.pattern_title : "<span class='label label-warning'>Not Answered </span>".html_safe %></td>
-                <td><%= (thought.challenging_thought && thought.challenging_thought.length > 0) ? thought.challenging_thought : "<span class='label label-warning'>Not Answered </span>".html_safe %></td>
-                <td><%= (thought.act_as_if && thought.act_as_if.length > 0) ? thought.act_as_if : "<span class='label label-warning'>Not Answered </span>".html_safe %></td>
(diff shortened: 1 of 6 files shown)


app/views/think_feel_do_engine/coach/group_dashboard/_thoughts.html.erb
```diff
@@ -31,7 +33,9 @@
                 </td>
                 <td><%= membership.participant.display_name %></td>
                 <td><%= thought.content %></td>
-                <td><%= thought.effect %></td>
+                <td><%= (thought.pattern_title && thought.pattern_title.length > 0) ? thought.pattern_title : "<span class='label label-warning'>Not Answered </span>".html_safe %></td>
+                <td><%= (thought.challenging_thought && thought.challenging_thought.length > 0) ? thought.challenging_thought : "<span class='label label-warning'>Not Answered </span>".html_safe %></td>
+                <td><%= (thought.act_as_if && thought.act_as_if.length > 0) ? thought.act_as_if : "<span class='label label-warning'>Not Answered </span>".html_safe %></td>
```
michaelwehrley: In the future, to shorten up, you could call ```empty?``` on the string instead of ```length ? 0```  Also, because these three lines #L36-#38 all have the same logic, is it reasonable to create a helper and pass in the attribute name?  In that case, the helper's logic could also be tested.  Oh, and one last thing, I believe ```&&``` is quite high in the order of operations chain, so you most likely don't need parentheses.

For example, something like this:
```
def display_answer_status(thought, attr)
  thought.send(:attr) && thought.send(:attr).empty? ? thought.challenging_thought : content_tag(:span, "Not Answered", class: "label label-warning" )
end
```

michaelwehrley:
:+1:

app/views/think_feel_do_engine/coach/group_dashboard/_thoughts.html.erb
eschlange: I'll take care of the empty? switch and add a helper method. The parentheses, while not necessary, I added for readability. I can remove them though.

PR #95, opened 2015-03-06
Minor tweak to fix group dash activities  view

cbrenner04:


gwensmu:
a reasonable change ⛵

PR #98, opened 2015-03-09
Fix goal like count for group dash.

eschlange:
* Updated the group dash helper to include new method.
* Testing new method for goal like count in new spec test.
* Added dummy models in order to implement spec.
* Updated view to use the new helper method.

[Finished #89835632]

### DIFF
--- a/app/helpers/think_feel_do_engine/coach/group_dashboard_helper.rb
+++ b/app/helpers/think_feel_do_engine/coach/group_dashboard_helper.rb
@@ -116,6 +116,20 @@ module ThinkFeelDoEngine
         string_to_check && !string_to_check.empty?
       end
 
+      def goal_like_count(goal)
+        like_total = 0
+        goal_shared_items =
+          SocialNetworking::SharedItem
+          .where(item_type: "SocialNetworking::Goal", item_id: goal.id)
+        goal_shared_items.each do |item|
+          like_total +=
+            SocialNetworking::Like
+            .where(item_type: "SocialNetworking::SharedItem",
+                   item_id: item.id).count
+        end
+        like_total
+      end
+
       private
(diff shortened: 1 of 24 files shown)


app/helpers/think_feel_do_engine/coach/group_dashboard_helper.rb
```diff
@@ -112,6 +112,24 @@ def activity_status(activity)
         end
       end
 
+      def not_nil_and_populated_string(string_to_check)
```
gwensmu: So this excludes strings that are "", basically?

eschlange: That is correct.

app/views/think_feel_do_engine/coach/group_dashboard/_thoughts.html.erb
```diff
@@ -31,7 +33,9 @@
                 </td>
                 <td><%= membership.participant.display_name %></td>
                 <td><%= thought.content %></td>
-                <td><%= thought.effect %></td>
+                <td><%= not_nil_and_populated_string(thought.pattern_title) ? thought.pattern_title : "<span class='label label-warning'>Not Answered </span>".html_safe %></td>
```
gwensmu: use content_tag instead of html_safe?

eschlange: I'll fix that as a separate story, that line was added as a result of a merge conflict and is unrelated to this story.

gwensmu: ok cool! i'll merge this.

PR #99, opened 2015-03-09
Ecf 89823022 available content modules

ericcf:


app/models/tool_nav_item.rb
```diff
@@ -34,33 +34,22 @@ def has_subnav?
   end
 
   def module_nav_items
-    ModuleNavItem.for_content_modules(@participant, available_content_modules)
+    available_content_modules
```
gwensmu: so readable

app/models/tool_nav_item.rb
```diff
@@ -34,33 +34,22 @@ def has_subnav?
...
 
   def any_incomplete_tasks_today?
-    content_module_ids = available_content_modules.where(is_viz: false)
-                         .map(&:id)
-
-    @participant.active_membership
-      .incomplete_tasks_today
-      .for_content_module_ids(content_module_ids)
-      .exists?
+    available_content_modules.is_not_viz.is_not_completed.exists?
```
gwensmu: i could weep!

app/views/think_feel_do_engine/participants/content_modules/index.html.erb
```diff
@@ -44,25 +43,20 @@
...
+        .new(self, nil, nil, current_participant),
+        navigator_location_path(
+          module_id: content_module.id
         )
-        .html_safe 
-        # render the viz provider
-    %>
-  <% end %>
+      )
+      .html_safe
```
gwensmu: content_tag instead of html_safe?

ericcf: I agree. I'll add a new commit.

ericcf: Actually, this is a bigger refactor than I thought. Not sure it's feasible right now without a pretty big rewrite.

gwensmu: ok. the content module titles aren't created via web anyway, they are imported direct to database, so this isn't as big of an issue as other uses, right?

ericcf: It's actually calling the `render_current` method on a `ContentProvider` subclass instance, e.g. https://github.com/cbitstech/think_feel_do_engine/blob/ecf_89823022_available_content_modules/app/models/content_providers/activity_index_provider.rb#L9

gwensmu:
🌅

PR #100, opened 2015-03-10
Modularize patient dashboard with partials

gwensmu:
* Move each discrete table or section into a partial
* Move javascript to handle tables into separate file
  under /assets
* These changes will allow us to customize the dashboards
  in host apps more easily.
* Move PHQ9 table out of engine (into host app)
* Move PHQ9 table spec out of engine (into host app)

[#89312806]


eschlange:
:octopus:  Looks good! I'll merge it when the build completes with success.

PR #103, opened 2015-03-10
refactor Group methods with Ruby conventions

ericcf:


michaelwehrley:
:+1: :cake:

app/models/group.rb
```diff
@@ -29,133 +29,109 @@ def learning_tasks
...
         .participant
         .participant_login_events
         .where(participant_login_events[:created_at]
                  .gteq(week_start_day(week_number)))
         .where(participant_login_events[:created_at]
                  .lt(week_end_day(week_number)))
-        .count
-    end
-    login_count
+    end.map(&:count).sum
```
eschlange: Yes! I looked for this sort of solution for quite awhile. I'm finally getting to understanding procs and blocks better. Does passing a symbol as a proc into a method always call that proc at the end of whatever block is being run?

ericcf: This probably explains it better than I could: http://stackoverflow.com/questions/1961030/ruby-ampersand-colon-shortcut#answer-1961118. Basically, any method that accepts a block and yields a single argument can use the `&` shortcut.

PR #107, opened 2015-03-11
Reporter sanitizes the file path more strictly

gwensmu:
* Only allow [a-z]+ charecters
* Discard the rest
* Add a spec to test cleanliness

[#90118312]


app/models/think_feel_do_engine/reports/reporter.rb
```diff
@@ -14,7 +14,7 @@ def self.fetch_reports
 
       def self.file_path(name)
         Rails.application.config.reports_root_path
-          .join("reports/#{name.to_s.gsub('_', '').downcase}.csv")
+          .join("reports/#{name.to_s.scan(/[a-z]+/).join.downcase}.csv")
```
ericcf: Nice!

PR #108, opened 2015-03-12
Add like table to group dash.

eschlange:
* Add like table to group dash view.
* Add link to like table at top of dash view.

[Finished #90054316]

### DIFF
--- a/app/helpers/think_feel_do_engine/coach/group_dashboard_helper.rb
+++ b/app/helpers/think_feel_do_engine/coach/group_dashboard_helper.rb
@@ -58,10 +58,6 @@ module ThinkFeelDoEngine
         end
       end
 
-      def like_item_description(like)
-        "some description"
-      end
-
       def participants_that_read_lesson(task)
         render partial: "think_feel_do_engine/coach/group_dashboard/"\
                         "lesson_completion_breakdown",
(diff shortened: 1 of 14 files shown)


app/helpers/think_feel_do_engine/coach/group_dashboard_helper.rb
```diff
@@ -108,6 +108,10 @@ def membership_comments(membership)
         SocialNetworking::Comment.where(participant: membership.participant)
       end
 
+      def membership_likes(membership)
+        SocialNetworking::Like.where(participant: membership.participant)
```
ericcf: Would this be faster if you used `where(participant_id: membership.participant_id)` instead, or is it about the same?

ericcf:
@eschlange is there a feature spec somewhere to confirm that the likes table is included?

app/helpers/think_feel_do_engine/coach/group_dashboard_helper.rb
eschlange: Locally, its virtually the same. It is averaging a few milliseconds apart.

eschlange:
@ericcf The last push includes a feature spec to check for the existence of the likes container.

ericcf:
@eschlange cool, thanks! Will merge unless RuboCop rears his ugly head. :cop:

PR #109, opened 2015-03-13
Move reports from TFDSo host app into Engine

gwensmu:
* Reports for thought, activity and emotional rating
* Add specs for reports
* Set app config variables in dashboard
  specs so as to not conflict

[#90296538]

### DIFF
--- a/spec/models/think_feel_do_engine/reports/activity_spec.rb
+++ b/spec/models/think_feel_do_engine/reports/activity_spec.rb
@@ -1,0 +1,33 @@
+require "rails_helper"
+
+module ThinkFeelDoEngine
+  module Reports
+    RSpec.describe PatientActivity do
+      fixtures :all
+
+      describe ".all" do
+        context "when no thoughts recorded" do
+          it "returns an empty array" do
+            Activity.destroy_all
+
+            expect(PatientActivity.all).to be_empty
+          end
+        end
+
+        context "when thoughts recorded" do
+          it "returns accurate summaries" do
+            data = PatientActivity.all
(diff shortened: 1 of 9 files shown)


spec/models/think_feel_do_engine/reports/activity_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+            expect(PatientActivity.all).to be_empty
+          end
+        end
+
+        context "when thoughts recorded" do
+          it "returns accurate summaries" do
+            data = PatientActivity.all
+            activity = activities(:planned_activity1)
+            participant = participants(:participant1)
+            expect(data.count).to eq 18
```
michaelwehrley: Is this brittle?  What is the rationale for it being 18?

spec/models/think_feel_do_engine/reports/emotional_rating_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+            expect(EmotionalRating.all).to be_empty
+          end
+        end
+
+        context "when thoughts recorded" do
+          it "returns accurate summaries" do
+            data = EmotionalRating.all
+            emotional_rating = emotional_ratings(:longing)
+            participant = participants(:participant1)
+            expect(data.count).to eq 6
```
michaelwehrley: Again, do we need to check the count here?  Can we check that the count is not 0 for instance?

michaelwehrley:
:+1: After talking to @ericcf, testing count seems reasonable because the ```count``` won't always match ```all```.

PR #116, opened 2015-03-18
Update group dash goal view.

eschlange:
* Updated view for group dash.
* Goals now include dates for deleted and completed.

[Finished #90620628]


michaelwehrley:
:+1: :fireworks:

PR #117, opened 2015-03-19
Added wait time to test.

eschlange:
* Added delay to reduce CI slowdown error on test case.

[Finished #90512934]

### DIFF
--- a/spec/features/coach/patient_dashboard_spec.rb
+++ b/spec/features/coach/patient_dashboard_spec.rb
@@ -460,9 +460,11 @@ feature "patient dashboard", type: :feature do
 
       expect(page).to have_text "Participant was successfully stepped."
 
-      within "#patient-#{participants(:participant1).id}" do
-        expect(page).to have_text "Stepped"
-        expect(page).to_not have_button "Step"
+      using_wait_time 10 do
+        within "#patient-#{participants(:participant1).id}" do
+          expect(page).to have_text "Stepped"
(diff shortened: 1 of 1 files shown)


gwensmu:
🏫

michaelwehrley:
@eschlange @gwensmu I don't believe that this is an ajax call - nor does the test require js.  I think another alternative is to just revisit the page.

```ruby
visit "/coach/groups/#{group1.id}/patient_dashboards"
within "#patient-#{participants(:participant1).id}" do
  expect(page).to have_text "Stepped"
  expect(page).to_not have_button "Step"
end
```

eschlange:
@michaelwehrley If this is not a JS delay issue, then is it a page load time issue? I'm not sure how your solution differs from what currently exists.

michaelwehrley:
@eschlange Good question.  I am not sure exactly.  I have had success with the ```visit``` method in the past for situations like this.  After some investigating, my thought is the capybara ```visit``` method makes sure that all requests are made and completed before looking and giving up.

https://robots.thoughtbot.com/write-reliable-asynchronous-integration-tests-with-capybara

eschlange:
@michaelwehrley Cool! I'll update the test case and run it a few times to see if it passes.

michaelwehrley:
@eschlange Thanks, but don't kill your self; it is just a thought - I am not exactly sure. :smile:

PR #119, opened 2015-03-19
Test for Timeout

michaelwehrley:
* Check for the name of the participant first.


eschlange:
@michaelwehrley Wasn't this causing timeouts earlier today?

michaelwehrley:
@eschlange I was just so darn perplexed I had to try for a moment.  In the test, instead I waited for content - not just css to visible.  I am not sure if we want to proceed this way given the work you put in, but I thought it was worth taking another stab.  Thanks!

eschlange:
Ah ok, maybe you can cycle the build a few more times to ensure that it doesn't fail, before we merge with master.

ericcf:
@eschlange @michaelwehrley can this branch be deleted?

PR #120, opened 2015-03-19
Removed two N+1 issues from sign in page.

eschlange:
* Added includes for a participant activegroup
* Added nested includes for the moderator of the group.

[#90408368]


gwensmu:
interesting good to see this example ⛵

PR #121, opened 2015-03-20
Add includes for pattern on participant thoughts.

eschlange:
* Solves the N+1 problem on the Thoughts Home page.

[Finished #90408368]


gwensmu:
nice 🏫

PR #122, opened 2015-03-20
Remove N+1 issue.

eschlange:
* Included emotion into the emotional_rating scope of participant.

[#90408368]


gwensmu:
another down 🐔

PR #123, opened 2015-03-20
Fix N+1 issue with message queries.

eschlange:
* Updated participant to include sender in received messages.

[#90408368]


gwensmu:
tracing them one by one i support it 📞

PR #124, opened 2015-03-20
Add includes for activity type on activity

gwensmu:
* Resolves N + 1 query alert from bullet gem
  on load of "DO Home"

[#89800140]


eschlange:
N+1-mania!

PR #125, opened 2015-03-23
Remove redundant javascript to solve js error.

eschlange:
* Removed identical datatable initializations.
* Centralized the call on the page the specifies the partials.

[Finished #90711482]


gwensmu:
great 🚂

PR #127, opened 2015-03-26
Update to patient_dashboard spec

cbrenner04:
* added another scope to example "allows a coach to step a participant"


eschlange:
:+1:

PR #129, opened 2015-03-31
Remove Discontinue button.

eschlange:
* Removed discontinue button from patient dash page.
* Button removed for non-stepped care host applications.

[#91284340]


michaelwehrley:
:+1: :rooster:

PR #130, opened 2015-04-01
Update planned activity calculations

eschlange:
* added helper methods for planned activity counts.
* added specs to test validity of planned activity counts.

[Finished #91421272]


michaelwehrley:
:+1: :honey_pot:

PR #131, opened 2015-04-02
Updated planned activity count

eschlange:
* Now using created_at to scope planned row counts.

[#91421272]


ericcf:
Seems reasonable...

PR #133, opened 2015-04-06
replace actual config mutations with method stubs

ericcf:
* this is an attempt to fix the sporadically failing spec at
  features/coach/patient_dashboard_spec.rb:452


eschlange:
:+1:

PR #135, opened 2015-04-08
Fix instance where method can be called on nil object.

eschlange:
* Updated the video session model to run a nil check before checking kind.

[Finished #92040770]

### DIFF
--- a/app/models/think_feel_do_engine/reports/video_session.rb
+++ b/app/models/think_feel_do_engine/reports/video_session.rb
@@ -51,7 +51,7 @@ module ThinkFeelDoEngine
                      .limit(1)
                      .first
 
-        return stop_event if stop_event.kind != "videoPause"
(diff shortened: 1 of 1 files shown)


app/models/think_feel_do_engine/reports/video_session.rb
```diff
@@ -51,7 +51,7 @@ def self.next_event(event)
                      .limit(1)
                      .first
 
-        return stop_event if stop_event.kind != "videoPause"
+        return stop_event if stop_event && stop_event.kind != "videoPause"
```
michaelwehrley: I think you could do `stop_event.try(:kind)`.

In addition, what does order of operations say here?  I believe `&&` is petty high up there?
In this case, is what you wrote actually `((stop_event && stop_event.kind) != "videoPause")`  ?

PR #138, opened 2015-04-10
Add message display logic based on new arm flag.

eschlange:
* Messaging pages now take arm flag into consideration.
* Updated arm model.
* Updated navigator logic to handle complete memberships.
* Added migrations for new flag on arm.

[#91448610]

* Messaging is now disabled for discontinued participants with the right arm settings.

### DIFF
--- a/app/controllers/think_feel_do_engine/navigator_controller.rb
+++ b/app/controllers/think_feel_do_engine/navigator_controller.rb
@@ -101,9 +101,7 @@ module ThinkFeelDoEngine
     end
 
     def home_tool
-      current_participant
-        .active_membership
-        .group
+      current_participant.current_group
         .arm
         .bit_core_tools.find_by_type("Tools::Home")
     end
--- a/app/controllers/think_feel_do_engine/navigator_controller.rb
+++ b/app/controllers/think_feel_do_engine/navigator_controller.rb
@@ -101,7 +101,9 @@ module ThinkFeelDoEngine
     end
 
     def home_tool
-      current_participant.current_group
+      current_participant
+        .active_membership
+        .group
         .arm
         .bit_core_tools.find_by_type("Tools::Home")
     end
--- a/app/models/arm.rb
+++ b/app/models/arm.rb
@@ -13,6 +13,7 @@ class Arm < ActiveRecord::Base
            dependent: :destroy
 
   validates :title, presence: true
+  validates :can_message_after_membership_complete, presence: true
 
   def display_name_required_for_membership?(participant, display_name)
     if social? && display_name.empty?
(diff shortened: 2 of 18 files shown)


app/controllers/think_feel_do_engine/navigator_controller.rb
```diff
@@ -101,7 +101,9 @@ def provider_exist?
     end
 
     def home_tool
-      current_participant.current_group
+      current_participant
```
ericcf: @eschlange can you explain the purpose for this replacement?

app/models/arm.rb
```diff
@@ -36,6 +36,10 @@ def woz?
     has_woz
   end
 
+  def messaging_after_membership_complete?
```
ericcf: Rather than adding a new method, what do you think about using either the automatically added `can_message_after_membership_complete?` or `can_message_after_membership_complete.present?`

db/migrate/20150410174157_add_can_message_after_membership_complete_to_arms.rb
```diff
@@ -0,0 +1,5 @@
+class AddCanMessageAfterMembershipCompleteToArms < ActiveRecord::Migration
+  def change
+    add_column :arms, :can_message_after_membership_complete, :boolean, null: false, default: true
```
ericcf: Should there be a corresponding model validation?

spec/features/participant/messages_spec.rb
```diff
@@ -127,3 +127,16 @@
...
+feature "discontinued participants can continue messaging if arm settings allow it" do
+  fixtures :all
+
+  before do
+    sign_in_participant participants(:participant_study_complete_2)
+    visit "/navigator/contexts/MESSAGES"
+  end
+
+  it "should have a compose button on the message home" do
+    expect(page).to_not have_button("Compose")
```
ericcf: Should there be a "Compose" button present?

app/controllers/think_feel_do_engine/navigator_controller.rb
eschlange: Good catch, my change was identical to the scoping defined through "current_group". This was added to due to caching issues on my local environment and I forgot to revert the change. I'll revert it when fixing the other issues mentioned in your review.

app/models/arm.rb
eschlange: I will make this change, old Java habits die hard. I'll need to update the dashboard and host apps as well for this change.

db/migrate/20150410174157_add_can_message_after_membership_complete_to_arms.rb
eschlange: Yes, I'll add in the validation now.

eschlange:
@ericcf @michaelwehrley Please look over my changes when you have a chance. I'll need this pull request merged  in before I can update the host applications for this feature.

ericcf:
:rotating_light:

PR #139, opened 2015-04-14
Update date-time format for patient and group dash.

eschlange:
* Updated date and time formats for activites future view.
* Updated date and time formats for audio access view.
* Updated date and time formats for inactive patients view.

[#91544146]


ericcf:
:yes:

PR #140, opened 2015-04-15
Update Arm validation for existing migration compatibility.

eschlange:
* Added custom validator for arms.

[#91448610]


ericcf:
Looks reasonable.

PR #141, opened 2015-04-16
make all report date(time)s use iso8601

ericcf:
[#90707874]

### DIFF
--- a/app/models/think_feel_do_engine/reports/video_session.rb
+++ b/app/models/think_feel_do_engine/reports/video_session.rb
@@ -24,8 +24,8 @@ module ThinkFeelDoEngine
             {
               participant_id: participant.study_id,
               video_title: video.try(:title),
-              video_started_at: e.emitted_at,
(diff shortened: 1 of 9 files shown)


app/models/think_feel_do_engine/reports/video_session.rb
```diff
@@ -24,8 +24,8 @@ def self.all
             {
               participant_id: participant.study_id,
               video_title: video.try(:title),
-              video_started_at: e.emitted_at,
-              video_stopped_at: next_e.try(:emitted_at),
+              video_started_at: e.emitted_at.iso8601,
+              video_stopped_at: next_e.try(:emitted_at).iso8601,
```
michaelwehrley: do you then have to do `try(:iso8601)`

ericcf: good catch! :fish:

michaelwehrley:
:+1: :eyeglasses: good when ready.

PR #145, opened 2015-04-22
Add datetime to site messages view

cbrenner04:
* Add datetime to messages view
* Update site_messages_spec.rb to test for new datetime


ericcf:
Looks :cool:

PR #146, opened 2015-04-22
Update table view in patient dashboard

cbrenner04:
* Update table view in patient dashboard to not show Stepped on Date when app
  has social features
* Add spec to test above update


cbrenner04:
closed because these commits are included in #147

PR #148, opened 2015-04-22
Update Mood/Emotions viz to conform to datetime

cbrenner04:


app/assets/javascripts/think_feel_do_engine/feel/emotions.js
```diff
@@ -406,10 +406,10 @@ function dailyDrillModal (data) {
   var charge = data[4] === "Emotion" ? (data[2] ? "Positive" : "Negative") : "";
   html = "";
   html += "<div class='modal fade' id='smallModal-"+guid+"' tabindex='-1' role='dialog' aria-labelledby='smallModal' aria-hidden='true'><div class='modal-dialog modal-sm'><div class='modal-content'><div class='modal-header'><button type='button' class='close' data-dismiss='modal' aria-hidden='true'>&times;</button>"
-  html += "<h4 class='modal-title' id='myModalLabel'><strong>"+charge+" "+data[4]+"</strong><br>"+ moment(new Date(data[0])).format('LL') + "</h4></div><div class='modal-body'>"
+  html += "<h4 class='modal-title' id='myModalLabel'><strong>"+charge+" "+data[4]+"</strong><br>"+ moment(new Date(data[0])).format('L') + "</h4></div><div class='modal-body'>"
   $.each(data[3], function(i, d){
     var emotion = d[2] == undefined ? '' : d[2]
-    html += "<p>"+moment(new Date(d[1])).format('hh:mm a')+": "+d[0]+" "+emotion+"</p>"
+    html += "<p>"+moment(new Date(d[1])).format('h:mm a')+": "+d[0]+" "+emotion+"</p>"
```
cbrenner04: I left this in a way that does not conform to datetime format standards because it would be redundant and difficult to understand in this particular instance

cbrenner04:
@eschlange I hard coded these date formats to the standard instead of using (format: :standard) because I was not sure how to implement that. If you could point me in the right direction I will update.

app/assets/javascripts/think_feel_do_engine/feel/emotions.js
eschlange: This is valid, since we can't utilize the ruby formatter for a javascript date.

eschlange:
:+1:  Looks good to me.

PR #150, opened 2015-04-22
Update faulty arm validator logic.

eschlange:
* Logic previously would not allow false for flag.

[#91448610]


ericcf:
I'm not at my computer, so please go ahead and merge. Looks good.

PR #152, opened 2015-04-23
Update date on x-axis in mood and emotions viz

cbrenner04:
* Change date so there isn't overlap of dates on x-axis in mobile view


michaelwehrley:
Do you also want to change it here: https://github.com/cbitstech/think_feel_do_engine/blob/cjb_emotions_viz/app/assets/javascripts/think_feel_do_engine/feel/emotions.js#L92

cbrenner04:
@michaelwehrley I am not sure. That date_format doesn't seem to have an effect on anything and given the standard for date is %m/%d/%Y, I'd rather just change the specific one for the x-axis and leave the rest to the standard.

eschlange:
Solid :rocket:

PR #153, opened 2015-04-23
Update datetime in coach/recieved_messages partial

cbrenner04:
* Update datetime in partial - spelling standard correctly
* Update coach/messages_spec.rb to reflect updates


eschlange:
Lookin' good~ :sunglasses:

PR #154, opened 2015-05-01
add missing tool description to yieldable content

ericcf:
* fix CSS issue with modal backdrop
* fix bug in view backed model association
* fix bug in view

[#90938810]


eschlange:
Is the modal backdrop problem introduced through Bootswatch themes, or is it a known issue with vanilla Bootstrap?

eschlange:
:+1: Looks good! I like the view spec, very concise.

PR #159, opened 2015-05-14
add blank predicted values for planned activities

ericcf:
* require predicted values

[#94335536]


app/views/think_feel_do_engine/activities/_build_activity.html.erb
```diff
@@ -1,12 +1,14 @@
+<%
+  parameter_prefix ||= "activity"
```
eschlange: I have not seen this syntax before, but I love it!

eschlange:
Looks good. :100:

PR #165, opened 2015-06-04
Update Snap to Media Query

michaelwehrley:
Update Snap to Media Query

* This change allows the snap file to exist but based on media queries
* Previously, the snap file displayed if the request was from a phone

[#95745360]


michaelwehrley:
Whomever looks at this, please also look at https://github.com/cbitstech/think_feel_do_so/pull/97

ericcf:
Seems reasonable. :whale2:

PR #166, opened 2015-06-09
Scope Site Messages

michaelwehrley:
Scope Site Messages

* Add `User` model method to scope participants to a user based on group.
* Add scoping to site_messaging controller actions index and new.
* Add controller and model specs.

[Finished: #92731296] [Finished: #92731434]

### DIFF
--- a/app/controllers/think_feel_do_engine/coach/site_messages_controller.rb
+++ b/app/controllers/think_feel_do_engine/coach/site_messages_controller.rb
@@ -6,8 +6,8 @@ module ThinkFeelDoEngine
       load_and_authorize_resource except: [:index]
 
       def index
-        authorize! :show, SiteMessage
-        participant_ids = @group.participant_ids
+        authorize! :index, SiteMessage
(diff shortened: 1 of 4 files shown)


app/controllers/think_feel_do_engine/coach/site_messages_controller.rb
```diff
@@ -6,8 +6,8 @@ class SiteMessagesController < ApplicationController
       load_and_authorize_resource except: [:index]
 
       def index
-        authorize! :show, SiteMessage
-        participant_ids = @group.participant_ids
+        authorize! :index, SiteMessage
+        participant_ids = current_user.participants_for_group(@group).try(:ids)
```
ericcf: Is `ids` a method on the Participant relation?

michaelwehrley: It is on the array of participants.  If there are now participants, then it would be a method on an empty array.

PR #168, opened 2015-06-11
Update Viz JS

michaelwehrley:
Update Viz JS
* Prevent Default on click event.

[#95718670]


michaelwehrley:
The test is in think_feel_do_so

PR #171, opened 2015-06-12
Update Week Testing Becomes Valid

michaelwehrley:
Update Week Testing Becomes Valid

* Change week 4 to week 5.
* Update specs with new constants.

[Finishes: #96449738]


ericcf:
Nice! I like replacing magic constants! :horse_racing:

michaelwehrley:
@ericcf I know you do! :smile:

PR #173, opened 2015-06-16
Improve Validations

michaelwehrley:
Improve Activity Form Validations

* User can't submit data if form is incomplete.
* Refactor for validations
* Add feature test for testing of validations and alert.

[Finished: #94683542]


eschlange:
Worked through the code with Mike and talk about refactoring solutions for activity CRUD. This is a valid solution until we need to do an overhaul of activity editing.

PR #176, opened 2015-06-17
Add time, date, and datetime format for login info

cbrenner04:
* This is specifically for the Patient Info and Login Info tables in pt dash

[Finished: #97202480]


michaelwehrley:
:+1: :pizza:

PR #177, opened 2015-06-18
Add format for time, update lesson view table

cbrenner04:
* Add a format for time and datetime that has seconds
* Update time formats in lesson view table on patient dashboard to show seconds
* Add column on lesson view table on patient dashboard to show duration

[Finish #96783814]


eschlange:
:+1:

PR #178, opened 2015-06-18
Make messages table filter and sortable

cbrenner04:


app/views/think_feel_do_engine/coach/patient_dashboards/tables/_messages_table.html.erb
```diff
@@ -11,9 +11,10 @@
   <div id="messagesCollapse" class="panel-collapse collapse in">
     <div class="panel-body">
       <div class="table-responsive div-table-viz">
-        <table class="table table-hover generic-data-table data-table" id="messages">
+        <table class="table table-hover data-table" id="messages">
           <thead>
           <tr>
+            <th class="not displayed"></th>
```
eschlange: The class for the initial table header should probably be "not-displayed" instead of "not displayed"

app/views/think_feel_do_engine/messages/_message.html.erb
```diff
@@ -1,4 +1,7 @@
 <tr>
+  <td class="not displayed">
```
eschlange: Same here: "not-displayed".

cbrenner04:
@eschlange crap. Thanks! I will make the edits and push again

PR #179, opened 2015-06-18
Fix Learning Table Template Error

michaelwehrley:
Fix Learning Table Template Error

* Check to see if a lesson event is nil (once flattened)
* Display only non-nil events in learning table

[Finished: #97251532]


michaelwehrley:
@eschlange I nested the tests so that is why it looks like there are so many new changes.  I actually only added the last test on line 44

eschlange:
:eyes: :+1:

PR #185, opened 2015-06-22
Update planning activity examples in specs

cbrenner04:
* Change `from:` to use css name instead of the label

[#97428920]


eschlange:
:pizza:

PR #189, opened 2015-06-25
Fix Timing Bug In Spec

michaelwehrley:
Fix Timing Bug In Spec

* Implement an after callback to reset timecop to rails_helper specifications.

[Finished: #97785500]


michaelwehrley:
@ericcf Because I run `Timecop.return` within the test (so the browser and the database have the same time environment) I need to reset Timecop to be what we set it in rails_helper.  I was thinking of editing the callback in rails_helper, but I saw we called `Timecop.return` there, and I thought if the whole testing suite isn't broken, don't fix it.

ericcf:
My suggestion is to use Timecop.travel with a block instead. I think it's really confusing to follow the logic when it's used this way.

michaelwehrley:
@ericcf we are good to go now.  Thanks!

PR #190, opened 2015-06-26
Fix Click Bug for Emotions Data Graph

michaelwehrley:
Fix Click Bug for Emotions Data Graph

* Change `change` event to `click`.
* Add jQuery - Jasmine specs.

[Finished: #97645366]


eschlange:
Looking good. :sunglasses:

PR #191, opened 2015-06-26
Update When ActiveMemebership is Checked

michaelwehrley:
Update When Active Membership is Checked

* Upon any request, check if participant is active.
* If `active_membership` is `nil`, the participant is logged out.

[Finished: #95530436]

### DIFF
--- a/app/controllers/think_feel_do_engine/participants/sessions_controller.rb
+++ b/app/controllers/think_feel_do_engine/participants/sessions_controller.rb
@@ -7,15 +7,8 @@ module ThinkFeelDoEngine
       skip_authorization_check
 
       def create
-        participant = Participant.find_by_email params[:participant][:email]
-        if participant && !participant.active_membership
-          msg = "We're sorry, but you can't sign in yet because you are not " \
-                "assigned to a group."
-          redirect_to new_participant_session_path, alert: msg
-        else
-          super do |resource|
-            ParticipantLoginEvent.create(participant_id: resource.id)
-          end
+        super do |resource|
+          ParticipantLoginEvent.create(participant_id: resource.id)
         end
       end
     end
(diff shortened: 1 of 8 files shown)


app/controllers/think_feel_do_engine/participants/sessions_controller.rb
```diff
@@ -9,9 +9,7 @@ class SessionsController < Devise::SessionsController
       def create
         participant = Participant.find_by_email params[:participant][:email]
         if participant && !participant.active_membership
```
michaelwehrley: @ericcf This check may now be redundant.

ericcf: Yes, I'd say remove it to simplify.

spec/controllers/think_feel_do_engine/participant_data_controller_spec.rb
```diff
@@ -13,7 +13,7 @@ module ThinkFeelDoEngine
     end
     let(:navigator) { instance_double("Navigator", current_content_provider: provider) }
     let(:data_record) { double("data_record") }
-    let(:participant) { double("participant", navigation_status: nil) }
+    let(:participant) { double("participant", active_membership: true, navigation_status: nil) }
```
ericcf: What do you think about using a verifying double here? https://relishapp.com/rspec/rspec-mocks/v/3-3/docs/verifying-doubles

app/controllers/think_feel_do_engine/participants/sessions_controller.rb
michaelwehrley: great!

spec/controllers/think_feel_do_engine/participant_data_controller_spec.rb
michaelwehrley: Are you thinking that if we say `instance_double(Participant, active_membership: true, navigation_status: nil)`, that if we change the attribute from `active_membership` to something else in the future, it will help us be aware b/c we will get an implementation error?  Or what made you think of this?  Thanks!

ericcf: Yes, if we change the name of or remove a method, any specs using its instance double will tell us immediately. Otherwise, the tests would continue to pass and give a false positive.

michaelwehrley: :awesome!  Thanks!

ericcf:
:koala:

PR #194, opened 2015-07-01
Fix Test Summary

michaelwehrley:
Fix PHQ Test Summary

* Remove attribute.
* Update spec for defined attribute.
* Added helper spec.
* Update feature spec to make sure phq test lable is displayed

[Finished: #98120678] [#97581666]


ericcf:
@michaelwehrley I think this is ok, but can you be a little more descriptive in your commit comments? E.g. "Remove attribute." - which attribute? Why is it being removed? Thanks!

michaelwehrley:
Good point.  The attribute `prev_range_start` is being removed b/c I forgot to remove it when I removed the method using it in a previous commit: 97581666.  Thanks!

PR #195, opened 2015-07-02
Remove New Alert from Tools

michaelwehrley:
Remove New Alert from Tools

* Conditionally load 'New!' for tools
* Support and Relax tools no longer have alert.
* Add model specs to check for conditions.

[Finished: #97468932]

### DIFF
--- a/spec/models/tool_nav_item_spec.rb
+++ b/spec/models/tool_nav_item_spec.rb
@@ -47,6 +47,41 @@ describe ToolNavItem do
         end
       end
     end
+
+    context "Support tools exist" do
+      let(:support) { bit_core_tools :support }
+      let(:nav_item) { ToolNavItem.new(participant, support) }
+
+      before do
+        allow(nav_item).to receive(:any_incomplete_tasks_today?) { true }
+      end
+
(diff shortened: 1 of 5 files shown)


spec/models/tool_nav_item_spec.rb
```diff
@@ -47,6 +47,41 @@ def tool
...
+      it "Verifies the alert doesn't display for the relax tool" do
+        expect(nav.display_alert(relax)).to eq false
+      end
+
+      it "Verifies the alert doesn't display for the support tool" do
+        expect(nav.display_alert(support)).to eq false
+      end
+
+      it "Verifies the alert shows for other tools" do
+        expect(nav.display_alert(support)).to eq false
```
cbrenner04: @michaelwehrley what is the point of this example? it looks exactly the same as the previous. It also doesn't seem to be testing what the description says

michaelwehrley: Thanks, that is what happens when in a rush.

cbrenner04: No problem. Figured as much. That's why we do the pull requests.

PR #198, opened 2015-07-06
Cl 91858816 remove lesson title markdown revised

Wrenderer:
Remove Markdown from Lesson Titles
* Add function to remove markdown patterns
* Add spec to confirm function works as intended
Update Markdown Removal Method
* Remove method from previous commit, `markdown_free_title`
* Change `title` in group dashboard to `pretty_title`
* Update view spec to use `pretty_title`
* Add model spec for `pretty_title`
Remove Markdown in Manage Tasks dashboard
* Remove delegate for `title` in Task class
* Add `title` method to Task Class which will automatically defer to `pretty_title` for lesson modules
* Add spec to Task spec to test the new `title` method functionality
* Refactored part of 'edit_tasks.html.erb'

### DIFF
--- a/spec/models/content_modules/lesson_module_spec.rb
+++ b/spec/models/content_modules/lesson_module_spec.rb
@@ -1,0 +1,15 @@
+require "rails_helper"
+
+module ContentModules
+  describe LessonModule do
+    describe "render_current" do
+      fixtures(:all)
+
+      let(:markdown) { bit_core_content_modules(:lesson_with_markdown) }
+
+      it "should have markdown free titles in the dashboard view" do
+        expect(markdown.pretty_title).to match(/Lesson <strong>With<\/strong> Markdown/)
+      end
+    end
+  end
+end
--- a/spec/models/content_modules/lesson_module_spec.rb
+++ b/spec/models/content_modules/lesson_module_spec.rb
@@ -2,12 +2,12 @@ require "rails_helper"
 
 module ContentModules
   describe LessonModule do
-    describe "render_current" do
+    describe "#pretty_title" do
       fixtures(:all)
 
       let(:markdown) { bit_core_content_modules(:lesson_with_markdown) }
 
-      it "should have markdown free titles in the dashboard view" do
+      it "returns a MarkDown-free string" do
         expect(markdown.pretty_title).to match(/Lesson <strong>With<\/strong> Markdown/)
       end
     end
--- a/spec/models/task_spec.rb
+++ b/spec/models/task_spec.rb
@@ -8,6 +8,7 @@ describe Task do
   )
 
   let(:task1) { tasks(:task1) }
+  let(:task_learning7) { tasks(:task_learning7) }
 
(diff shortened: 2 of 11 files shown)


michaelwehrley:
See https://github.com/cbitstech/think_feel_do_engine/commit/c37c928c690f9b95bda53c6f80bce4df13f809e5#commitcomment-12025167 for previous comments on old PR.

spec/models/content_modules/lesson_module_spec.rb
```diff
@@ -0,0 +1,15 @@
+require "rails_helper"
+
+module ContentModules
+  describe LessonModule do
+    describe "render_current" do
+      fixtures(:all)
+
+      let(:markdown) { bit_core_content_modules(:lesson_with_markdown) }
+
+      it "should have markdown free titles in the dashboard view" do
```
ericcf: This example description is a little misleading. What do you think about "returns a MarkDown-free string"?

spec/models/content_modules/lesson_module_spec.rb
```diff
@@ -0,0 +1,15 @@
+require "rails_helper"
+
+module ContentModules
+  describe LessonModule do
+    describe "render_current" do
```
ericcf: The RSpec convention we're using when providing an example group for an instance method is to give it a `#` prefix. E.g. here you would use `describe "#pretty_title"`. This documentation may be helpful: https://github.com/cbitstech/guides/tree/master/testing/rspec

spec/views/think_feel_do_engine/coach/group_dashboard/_learning_lesson_row.html.erb_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  before do
+    allow(view).to receive_message_chain(:content_provider, :source_content) { lesson_module }
+    allow(lesson_module).to receive(:pretty_title) { "Pretty Markdown" }
+    stub_template "think_feel_do_engine/coach/group_dashboard/_lesson_completion_breakdown" => ""
+    allow(view).to receive(:task) { nil }
+  end
+
+  it "runs the method `pretty_title`" do
+    expect(lesson_module).to receive(:pretty_title) { true }
```
ericcf: Here you're testing an implementation, rather than a behavior. That makes the test brittle, because if the implementation changes the test will break even though the behavior still works as expected.

spec/models/task_spec.rb
```diff
@@ -51,6 +52,16 @@
       end
     end
 
+    context "when being listed" do
+      it "displays non-html formatted title for non-lesson content modules" do
+        expect(task1.title).to eq("#1 Awareness")
+      end
+
+      it "displays html formatted title for lesson modules" do
+        expect(task_learning7.title).to match(/<.*>Feel - Emotions Introduction<\/.*>/)
```
michaelwehrley: :+1:

ericcf:
Looks good!

PR #202, opened 2015-07-14
Fix markdown changing to visible html tags in group dashboard

Wrenderer:
* Change link display in "_assigned_task_cell.html.erb" so that `html_safe` strings are preserved in the view (this keeps html tags from appearing in the view)

[Finished: #91858816]


ericcf:
Looks good to me. Go ahead and merge once Travis passes!

PR #205, opened 2015-07-21
Fix Assigning Task Bug

michaelwehrley:
* Change before_save callback `check_if_valid_release_day` to allow tasks to be
  created for memberships that are part of a group containing memberships that
  have either been withdrawn and/or discontinued.
* Add scope `continuing` to find `Memberships` where `is_complete` is `false`.

[Finished: #99521860]

### DIFF
--- a/app/models/membership.rb
+++ b/app/models/membership.rb
@@ -49,6 +49,10 @@ class Membership < ActiveRecord::Base
     )
   }
 
+  scope :continuing, lambda {
+    where(is_complete: false)
+  }
+
   def available_task_statuses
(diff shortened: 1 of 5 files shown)


michaelwehrley:
Here is a note from @ericcf from a mislabeled branch: https://github.com/cbitstech/think_feel_do_engine/commit/55b743908ea228f374a8b0eb475e7468bd8d67b2#commitcomment-12264741.  The comment was addressed.

eschlange:
Looking good :sunglasses:

PR #206, opened 2015-07-21
Set Cookie From Header For Angular POSTS

michaelwehrley:
* The ACHIEVE page is being processed by the ThinkFeelDoEngine::NavigatorController
  When this happens the XSRF token was not set in the cookie (only for SocialNetworking Controllers).
  With no XSRF token set in cookie, the angular POST/PUTs could not proceed and thus the 422.
* Add the `set_csrf_cookie_for_ng` in the TFD Engine application controller and whala!

[Finished: #99525368]

### DIFF
--- a/app/controllers/think_feel_do_engine/application_controller.rb
+++ b/app/controllers/think_feel_do_engine/application_controller.rb
@@ -11,6 +11,7 @@ module ThinkFeelDoEngine
     protect_from_forgery with: :exception
 
     before_action :detect_browser, :verify_active_membership
+    after_action :set_csrf_cookie_for_ng
 
     layout "application"
 
@@ -49,6 +50,12 @@ module ThinkFeelDoEngine
       end
     end
 
+    protected
+
+    def verified_request?
+      super || form_authenticity_token == request.headers["X-XSRF-TOKEN"]
+    end
+
     private
 
     def phq_features?
@@ -71,6 +78,11 @@ module ThinkFeelDoEngine
         Devise.sign_out_all_scopes ? sign_out : sign_out(scope)
         redirect_to new_participant_session_path, alert: INACTIVE_MESSAGE
       end
+    end
+
+    def set_csrf_cookie_for_ng
+      return unless protect_against_forgery?
+      cookies["XSRF-TOKEN"] = form_authenticity_token
     end
   end
 end


app/controllers/think_feel_do_engine/application_controller.rb
```diff
@@ -49,6 +50,12 @@ def access_denied_resource_path
       end
     end
 
+    protected
+
+    def verified_request?
+      super || form_authenticity_token == request.headers["X-XSRF-TOKEN"]
```
ericcf: Where does this header come from?

ericcf:
Is there a reason you can't grab the token client side like this: https://github.com/rails/jquery-ujs/blob/master/src/rails.js#L59-L62

app/controllers/think_feel_do_engine/application_controller.rb
michaelwehrley: It looks like the requests is defined by rails when an instance of the controller is created: https://github.com/rails/rails/blob/10d0c43c7937f65b2fc6fb87a87b1023bbd85674/actionpack/lib/action_controller/metal/request_forgery_protection.rb#L138.  This header is the on defined in the head with the method `csrf_meta_tags`. In addition, this `verified_request?` is defined by Rails itself: https://github.com/rails/rails/blob/10d0c43c7937f65b2fc6fb87a87b1023bbd85674/actionpack/lib/action_controller/metal/request_forgery_protection.rb#L259-273

michaelwehrley: Here is more info on Angular setting the HEADER: https://docs.angularjs.org/api/ng/service/$http: it says:
Cross Site Request Forgery (XSRF) Protection
XSRF is a technique by which an unauthorized site can gain your user's private data. Angular provides a mechanism to counter XSRF. When performing XHR requests, the $http service reads a ***token from a cookie (by default, XSRF-TOKEN) and sets it as an HTTP header (X-XSRF-TOKEN)***. Since only JavaScript that runs on your domain could read the cookie, your server can be assured that the XHR came from JavaScript running on your domain. The header will not be set for cross-domain requests.

michaelwehrley:
Wouldn't we have to write `csrfToken` in every js post - or are you saying we could place it in the application.js file.  If that is the case, then what is difference?

ericcf:
I think this is fine for now. IMO it would be cleaner to do this all client side, but I'm going to merge it and we can address at a later date.

PR #207, opened 2015-07-21
Open Activity Summary Panel

michaelwehrley:
Open Activity Summary Panel

* Activity Summary Panel is open by default
* Add specs to make sure that string passed in for the  
  array argument is opened on default.

[Finished: #97469438]


app/assets/javascripts/think_feel_do_engine/activities/your_activities.js
```diff
@@ -1,16 +1,28 @@
...
+    sc.openOnDefault(id);
+  });
 };
 
 sc.collapsableIcon = function(anchorTag) {
+  var POUND_SIGN = '#';
+
   $(anchorTag).on('show.bs.collapse', function(event) {
-    $('[href^=#' + event.currentTarget.id + ']').find('.fa-caret-right').addClass('fa-caret-down').removeClass('fa-caret-right');
+    sc.openOnDefault(POUND_SIGN + event.currentTarget.id);
```
eschlange: Having a POUND_SIGN, constant is probably unnecessary. You may as well just keep the actual pound sign. If you insist on using a constant, I would name it something like ID_PREFIX instead.

eschlange:
Looks good! I'll merge after the build is complete.

PR #209, opened 2015-07-22
Add TurboLinks Ignore To Hamburger Links

michaelwehrley:
Add TurboLinks Ignore To Hamburger Links

* Add data attribute to links so the page gets reloaded.  
  In this case, the page was the ACHIEVE page which loads angular.  
  The angular was not loading with turbolinks.

[Finished: #99232384]


eschlange:
:de:

PR #211, opened 2015-07-23
Make "Introduction to ..." say the app name instead of just "ThinkFee…

Wrenderer:
…lDo"

 * Change literal "ThinkFeelDo" to `application_name` variable in view

[Finished: #97469994]

### DIFF
--- a/app/views/devise/sessions/new.html.erb
+++ b/app/views/devise/sessions/new.html.erb
@@ -41,7 +41,7 @@
 
 <% if (slideshow = SlideshowAnchor.fetch(:home_intro)) %>
   <br/>
-  <%= link_to "Introduction to ThinkFeelDo", ThinkFeelDoEngine::Engine.routes.url_helpers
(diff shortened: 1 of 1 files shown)


app/views/devise/sessions/new.html.erb
```diff
@@ -41,7 +41,7 @@
 
 <% if (slideshow = SlideshowAnchor.fetch(:home_intro)) %>
   <br/>
-  <%= link_to "Introduction to ThinkFeelDo", ThinkFeelDoEngine::Engine.routes.url_helpers
+  <%= link_to "Introduction to #{t(:application_name)}", ThinkFeelDoEngine::Engine.routes.url_helpers
```
ericcf: Where does the `application_name` value get set?

Wrenderer: "config\locales\en.yml"
`en:
  application_name: "Sunnyside"` for instance

PR #212, opened 2015-07-24
Update Last Activity Partial

michaelwehrley:
Update Last Activity Partial

* Move `#last_actiion_at` method to Participant model in ThinkFeelDoEngine.
* Update partial to display when the last activity occured.
* Update partial to display duration of last logged in session.
* Update and add feature specs to test conditions if events exist.
* Add model spec to test `#duration_of_last_session` & `#latest_action_at`

[#98851862]

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -14,9 +14,11 @@ gemspec
 # gem "debugger"
 
 gem "devise", "~> 3.2"
+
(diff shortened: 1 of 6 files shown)


eschlange:
Clean and concise.

PR #213, opened 2015-07-24
Add Migration To Add Null False to Columns

michaelwehrley:
Add Migration To Add Null False to Columns

* Add migration that verifies that captured events have  a `recorded_at` datetime.

[#98851862]


ericcf:
What do you think about adding this to the event_capture repo? It makes sense for any dependent of that library I think.

michaelwehrley:
We could totally do that.

PR #214, opened 2015-07-27
Update Event Capture

michaelwehrley:
Update Event Capture

* Include latest version of event capture 0.1.1
* This version no longer allows emitted_at or recorded_at to be null.

[#98851862]


eschlange:
Cleaning out defunct specs~ Nice!

PR #219, opened 2015-07-28
add guards on membership date validation

ericcf:
* when dates are nil, the validation should be handled by the
  presence validator

[##99708210]

### DIFF
--- a/spec/models/membership_spec.rb
+++ b/spec/models/membership_spec.rb
@@ -31,7 +31,7 @@ describe Membership do
         .to eq false
     end
 
-    it "prevents is_complete=true and an end_date in the future" do
+    it "does not allow is_complete=true and an end_date in the future" do
       expect(
         Membership.new(end_date: Date.today + 3.days, is_complete: true)
         .tap(&:valid?)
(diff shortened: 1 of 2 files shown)


spec/models/membership_spec.rb
```diff
@@ -50,6 +50,15 @@
         expect(existing_membership).to be_valid
       end
     end
+
+    it "does not allow a nil end date to be created" do
+      expect(Membership.new(end_date: nil).tap(&:valid?).errors[:end_date].length)
```
michaelwehrley: :smile:

michaelwehrley:
perfecto!

spec/models/membership_spec.rb
```diff
@@ -50,6 +50,15 @@
...
+        .to be > 0
+    end
+
+    it "does not allow a nil end date to be updated" do
+      expect(Membership.first.update(end_date: nil)).to eq false
```
michaelwehrley: I like this.  You don't have to check for error messages.  I really like this.  Dah!

PR #220, opened 2015-07-28
Remove Unnecessary try call

michaelwehrley:
Remove Unnecessary try call

* Remove unnecessary `try` call on event.

[Finished: #98485472]


michaelwehrley:
Note: https://github.com/cbitstech/think_feel_do_so/pull/120

PR #221, opened 2015-07-29
Add Try Method

michaelwehrley:
Add Try Method

* Add back in `try` method for determining the last event's `recorded_at`
* This is needed if no events exist.

[Finished: #98851862]


Wrenderer:
:+1: good to go if all the checks pass

PR #222, opened 2015-07-30
Change Duration Calculation

michaelwehrley:
* Wrong assumption was made about `last_sign_in_at`
* Using method `current_sign_in_at` on `Participant`
* `current_sign_in_at` and `last_sign_in_at` are the same if the participant  
  logs in for the first time.
* `current_sign_in_at` is the timestamp we want even if the participant is logged out
* `last_sign_in_at gets` updated when the participant logs back in.

[Finished: #98851862]

### DIFF
--- a/app/models/participant.rb
+++ b/app/models/participant.rb
@@ -103,7 +103,7 @@ class Participant < ActiveRecord::Base
   end
 
   def duration_of_last_session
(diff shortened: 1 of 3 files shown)


eschlange:
Those sign_in variable names are somewhat misleading. Looks good though.

eschlange:
I'll merge when the build is complete.

PR #227, opened 2015-08-10
Ecf 100846610 module session reports

ericcf:


app/models/think_feel_do_engine/reports/tool_module.rb
```diff
@@ -6,25 +6,17 @@ module ToolModule
 
       def self.included(klass)
         class << klass
```
eschlange: This method for adding methods to an instance is new to me!

eschlange:
Code compression + increased test coverage = :1, :2 :facepunch:

PR #229, opened 2015-08-11
Ecf 100969320 harmful thoughts table

ericcf:


ericcf:
Merging my own request.

PR #231, opened 2015-08-18
Show email name when display_name unavailable

Wrenderer:
* Add display_name method to participant model to handle blank display names

[Finished: #100974480]

### DIFF
--- a/app/models/participant.rb
+++ b/app/models/participant.rb
@@ -92,6 +92,14 @@ class Participant < ActiveRecord::Base
       )
   }
 
+  def display_name
+    if self[:display_name].blank?
+      self[:email].split("@").first
+    else
+      self[:display_name]
+    end
+  end
+
   def is_not_allowed_in_site
     # participant not set to is_complete (hence withdrawal or termination)
     # and who have no active memberships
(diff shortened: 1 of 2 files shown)


app/models/participant.rb
```diff
@@ -92,6 +92,14 @@ def self.inactive
       )
   }
 
+  def display_name
+    if self[:display_name].blank?
```
ericcf: You should be able to shorten this method body to just:

```ruby
self[:display_name] || email[/[^@]+/]
```

What do you think?

Wrenderer: I think I love one-line solutions! Unfortunately I just checked and this won't work. `display_name` is apparently a blank string by default and isn't allowed to be nill so that line will just output nothing when a display_name isn't defined - I'll update the commit comment to reflect that it can't actually be nill though it can be blank.

I originally had a line similar to:
`self[:display_name].blank? ? email[/[^@]+/] : self[:display_name]`

which does functionally work, but for readability reasons I changed it to the if/else. Thoughts?

ericcf: Ok, just leave as is. I'd like to see a couple of unit tests and then I think it's good to go. Thanks.

Wrenderer:
@ericcf Unit tests added

ericcf:
:tada:

PR #232, opened 2015-08-20
Update Jquery DataTables Columns

michaelwehrley:
Update jQuery DataTables Columns

* Change column for sorting to be created_at column.
* Move script tag for table within table.

[#101704324]


ericcf:
:octocat:

PR #234, opened 2015-08-24
Update Activity Time Formatting in New activity provider

michaelwehrley:
Update Activity Time Formatting

* Update Activity dateFormat in new activity provider (activity_tracker.js)
* Add Moment wrapper.
* Replace `moment()` with `momentjs()`.
* Name spaced `momentjs` to global `sc`.

[#91861818]


app/assets/javascripts/think_feel_do_engine.js
```diff
@@ -28,4 +28,5 @@
 //= require think_feel_do_engine/validate_social_sharing
 //= require think_feel_do_engine/markdown_editor.js
 //= require think_feel_do_engine/media_access_events.js
+//= require think_feel_do_engine/momentWrapper
```
michaelwehrley: @ericcf I wrote this as part of trying to improve the code within `activity_tracker.js`  Let me know your thoughts.  Thanks!

ericcf: One thought - call it a different name. One purpose in "wrapping" a service is to decouple the rest of your code from it. Naming the wrapper after the original service still couples it.

app/assets/javascripts/think_feel_do_engine/momentWrapper.js
```diff
@@ -0,0 +1,11 @@
+(function(momentFn) {
+  'use strict';
+
+  window.sc = window.sc || {};
+
+  function momentService(momentjs) {
+    return momentjs;
+  };
+
+  return window.sc.momentjs = momentService(momentFn);
```
ericcf: Why use a method call as opposed to just `window.sc.momentjs = momentFn;`? Also, why use a `return` statement?

app/assets/javascripts/think_feel_do_engine.js
michaelwehrley: Interesting point.  Thanks.

PR #237, opened 2015-08-25
ensure module session reports succeed

ericcf:
* even if for some reason a module is missing a provider, the report
  generation should still succeed

[#100846610]


michaelwehrley:
:eyeglasses: good!

PR #238, opened 2015-08-25
Msw 91861818 update time formats

michaelwehrley:


app/assets/javascripts/think_feel_do_engine/stepped_care.js
```diff
@@ -1,6 +1,6 @@
 (function() {
 
-window.sc = {};
+window.sc = window.sc || {};
```
ericcf: Is there a reason for initializing `window.sc` in multiple files?

michaelwehrley: good point.  i didn't want ordering of files to be a problem.

PR #240, opened 2015-08-26
Updated TimePicker Formatting

michaelwehrley:
Updated TimePicker Formatting

* Update TimePicker Formatting from default to 'M dd yy'
* Note: I created a function closure around activity_tracker.js.  
  While it seems like I changed a lot; I didn't.

[#91861818]


eschlange:
:pear:

PR #244, opened 2015-09-10
Remove Shareable Option

michaelwehrley:
* Remove option for sharing while creating activities.
* Add spec to check for the css of the sharing tag.

[#97469570]


eschlange:
Nice code reduction :+1:

PR #246, opened 2015-09-16
Add Default Week Count

michaelwehrley:
* When `study_length_in_weeks` is not defined, set default study length
* Update specs

[#102078078]


app/models/content_providers/learn_lessons_index_provider.rb
```diff
@@ -48,5 +48,13 @@ def content_modules(app_context)
         .where("bit_core_tools.title" => app_context)
         .where.not(id: bit_core_content_module_id)
     end
+
+    def week_count
+      Rails.application.config.study_length_in_weeks
```
eschlange: Ended up using rails app config, I see :eyes:

eschlange:
@michaelwehrley Are these build issues due to config issues, or mocking issues?

michaelwehrley:
@eschlange to answer your question, yes they were do to the Sentry references.  However, I removed those references and simplified the tests.

PR #247, opened 2015-09-21
Fix Duration Bug

michaelwehrley:
* Add correct parameters for displaying distance in words
* Update view to use presenter to remove logic from the view
* Add view specs and presenter specs.

[Finished: #102591154]


ericcf:
:shaved_ice:

PR #248, opened 2015-09-23
Fix CSS for In/Completed Modules

michaelwehrley:
* Update css to TFD Engine for completed or incomplete tasks
* Add helper class to build out task status link
* Add helper class in html

[Finished: #99263156]

### DIFF
--- a/app/assets/javascripts/think_feel_do_engine/task_status.js
+++ b/app/assets/javascripts/think_feel_do_engine/task_status.js
@@ -9,7 +9,6 @@
       $target = $(event.currentTarget);
       $taskStatusId = $target.data("task-status-id");
       $.ajax({
-        async: false,
         dataType: "script",
         type: "PUT",
         url: "/participants/task_status/" + $taskStatusId
--- a/app/assets/javascripts/think_feel_do_engine/task_status.js
+++ b/app/assets/javascripts/think_feel_do_engine/task_status.js
@@ -3,7 +3,7 @@
 
   // Report actions associated with task statuses
   $(document).on("page:change", function() {
-    $("a.task-status").not(".disabled").on("click", function(event) {
+\    $("a.task-status").not(".disabled").on("click", function(event) {
       var $target, $taskStatusId;
       
       $target = $(event.currentTarget);
--- a/app/assets/javascripts/think_feel_do_engine/task_status.js
+++ b/app/assets/javascripts/think_feel_do_engine/task_status.js
@@ -2,7 +2,7 @@
   "use strict";
 
   // Report actions associated with task statuses
-  $(document).on("ready, page:change", function() {
+  $(document).on("page:ready, page:change", function() {
     $("a.task-status").not(".disabled").on("click", function(event) {
       var $target, $taskStatusId;
       
--- a/app/assets/javascripts/think_feel_do_engine/task_status.js
+++ b/app/assets/javascripts/think_feel_do_engine/task_status.js
@@ -3,12 +3,13 @@
 
   // Report actions associated with task statuses
   $(document).on("page:change", function() {
-\    $("a.task-status").not(".disabled").on("click", function(event) {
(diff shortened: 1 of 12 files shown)


app/assets/javascripts/think_feel_do_engine/task_status.js
```diff
@@ -2,7 +2,7 @@
   "use strict";
 
   // Report actions associated with task statuses
-  $(document).on("ready, page:change", function() {
+  $(document).on("page:ready, page:change", function() {
```
ericcf: The documentation for jQuery says that the first parameter, `events`, should be space separated. I'm not sure it will work as expected with commas (I realize this was already here, just letting you know).

app/assets/stylesheets/think_feel_do_engine/base.css.scss
```diff
@@ -51,6 +51,12 @@ abbr.timeago {
 
 .list-group-item-unread {
   font-weight:bold;
+  background: #FFFFFF;
```
ericcf: Any reason not to just use `white` here?

app/assets/stylesheets/think_feel_do_engine/base.css.scss
```diff
@@ -51,6 +51,12 @@ abbr.timeago {
...
+}
+
+.list-group-item-read {
+  background: #FAFAFA;
```
ericcf: Since this is Scss, should this be assigned to a variable?

app/assets/javascripts/think_feel_do_engine/task_status.js
michaelwehrley: That does seem to be true: http://api.jquery.com/on/#on-events-selector-data-handler.  Thanks!  And good catch!

app/assets/stylesheets/think_feel_do_engine/base.css.scss
michaelwehrley: For readability?

app/assets/stylesheets/think_feel_do_engine/base.css.scss
michaelwehrley: could be. sure :+1:

app/helpers/think_feel_do_engine/tasks_helper.rb
```diff
@@ -2,6 +2,13 @@ module ThinkFeelDoEngine
   # Used to display asterisks if tasks and tools have been assigned to a group
   # and to hide unassigned links
   module TasksHelper
+    def available_module(available_module:, icon:, membership:)
```
ericcf: I think this method name is a little confusing given the first argument.

michaelwehrley:
@ericcf I was able to get this to pass and the problem was that I made an async call synchronous.  This is likely not what we want as task status updates can occur in the background.  We should probably address this in the future.  In addition, I changed the order that travis runs its rake tasks.  I would prefer to have the linting specs to run first as those run more quickly and if they fail we can then update sooner.  Thanks again for reviewing!

ericcf:
@michaelwehrley I think it's ok to make this a synchronous call as long as you don't notice a lag after clicking a link.

ericcf:
I like the idea of reordering the Travis tasks too!

ericcf:
I see you removed the listener on the `ready` event. Does this still work if you go directly to the URL and not navigate there via turbolinks?

michaelwehrley:
The handler was actually getting attached twice which is why I removed it. Not a huge deal, but it was making two `PUT` requests and I figured in other places, this could be a bigger deal.  I did manual testing, and the event always got called whether the link had turbolinks activated or not.  However, when I was trying to debug my failing specs on travis, I added the `ready` event back into the task_status.js method.  I removed it and pushed again.  Thanks!

ericcf:
Ok thanks!

PR #251, opened 2015-10-06
Remove Turbolink from Home

michaelwehrley:
* Add data attribute to ignore turbolink on home page

[Finished: #97469636]


ericcf:
Such turbo.

PR #252, opened 2015-10-07
Rescue ActionView::Template::Error

michaelwehrley:
* render view before `rescue` and within `rescue`.
* Update `rescue` to catch template error when group_id is not available.

[Finished: #101805508]


ericcf:
:cookie:

PR #254, opened 2015-10-08
Change Variable Name

michaelwehrley:
* Instance variable name was incorrectly `media_access_event_id`
* Changed `media_access_event_id` to match parameter: `mediaAccessEventId`.

[Finished: #102162512]


ericcf:
Nice catch!

PR #258, opened 2015-10-09
Update Sorting

michaelwehrley:
* Turn off sorting for final column in THINK.
* Update dataTable `columnDefs` key name based on jquery dataTable update.

[Finished: #101800918 ]


michaelwehrley:
@eschlange The big update is here: https://github.com/cbitstech/think_feel_do_engine/blob/msw_101800918_turn_off_sorting_for_column/app/views/think_feel_do_engine/thoughts/thoughts_table.html.erb#L53-L56

eschlange:
:nice-array-indexing

PR #259, opened 2015-10-09
Update SMS Notification Class

michaelwehrley:
* Refactor SMS notification and add spec
* Add host applications' lesson notification rake task
* Update rake task to consider app name.

[#105006670]
[#105229798]


app/models/delivered_message.rb
```diff
@@ -43,4 +47,13 @@ def deliver_notifications
   rescue
```
ericcf: This seems like a good spot to notify Sentry with Raven. What do you think?

app/models/message_sms_notification.rb
```diff
@@ -1,29 +1,44 @@
...
+  def initialize(body:, phone_number:)
+    @body = body
+    @phone_number = phone_number
   end
 
-  def self.deliver(message)
-    if Rails.env.staging? || Rails.env.production?
+  def deliver
+    if environment.staging? || environment.production?
       client = Twilio::REST::Client.new(
```
ericcf: Would this be worth extracting to the following?
```ruby
def sms_client
  @sms_client ||= Twilio::REST::Client.new(
    config.twilio_account_sid,
    config.twilio_auth_token
  )
end
```

app/models/message_sms_notification.rb
```diff
@@ -1,29 +1,44 @@
...
+    Rails.env
+  end
+
+  def from_telephone_number
+    config.try(:twilio_account_telephone_number)
+  end
+
+  def message
+    {
+      to: "+#{ phone_number }",
```
ericcf: Any reason not to prepend with `+1` so that clinicians just have to enter a 10 digit standard number?

michaelwehrley: good idea.  Thanks!

app/models/delivered_message.rb
michaelwehrley: :smile:

app/models/message_sms_notification.rb
michaelwehrley: Good catch.

app/models/message_sms_notification.rb
```diff
@@ -1,29 +1,47 @@
...
-      client = Twilio::REST::Client.new(
-        Rails.application.config.twilio_account_sid,
-        Rails.application.config.twilio_auth_token
-      )
+  def deliver
+    if environment.staging? || environment.production?
+      sms_client
 
-      client.account.messages.create message
+      @sms_client.account.messages.create message
```
ericcf: `sms_client.account.messages.create message`

app/models/message_sms_notification.rb
```diff
@@ -1,29 +1,47 @@
...
+    Rails.env
+  end
+
+  def from_telephone_number
+    config.try(:twilio_account_telephone_number)
+  end
+
+  def message
+    {
+      to: "+1#{ phone_number }",
```
ericcf: :+1:

PR #260, opened 2015-10-13
Add Password Concern

michaelwehrley:
* Move password validations to concern
* Password must contain integer
* Password must contain lowercase letter
* Password must contain uppercase letter
* Password can't include email.

[#98987112]


spec/models/think_feel_do_engine/concerns/validate_password_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+        Participant.create(
+          contact_preference: "email",
+          display_name: "H20",
+          email: "participant99@example.com",
+          password: password,
+          password_confirmation: password,
+          study_id: "TFD-9999")
+      end
+
+      def update_particiant(participant, password)
```
ericcf: Maybe rename this to `update_password` or `update_participant_password` for clarity?

spec/models/think_feel_do_engine/concerns/validate_password_spec.rb
```diff
@@ -0,0 +1,79 @@
+require "rails_helper"
+
+module ThinkFeelDoEngine
+  module Concerns
+    describe ValidatePassword do
+      let(:error_message) { "is too weak" }
+      let(:participant) do
```
ericcf: Do you ever validate this Participant + password? Maybe either explicitly call `.valid?` somewhere or use the `create!` method so that an exception is raised if it's not valid?

spec/models/think_feel_do_engine/concerns/validate_password_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+        password = "9c_$Fx^-!dk?/"
+        Participant.create!(
```
michaelwehrley: Not sure why this isn't showing up in github, but I used `create!`

ericcf:
Looks good to me!

PR #262, opened 2015-10-13
Add Validation Specs

michaelwehrley:
* Add specs to validate presence of `release day` and `participant`
* Add specs covering public methods

[#100127766]


eschlange:
Nice tests~

spec/models/phq_assessment_spec.rb
```diff
@@ -14,23 +14,106 @@
...
     context "when a response is above the maximum allowed" do
       it "is not valid" do
-        expect(PhqAssessment.new(valid_attributes.merge(q4: 4))).not_to be_valid
+        expect(PhqAssessment.new(valid_attributes.merge(q4: 4)))
+          .not_to be_valid
       end
     end
   end
+
+  describe ".completed?" do
```
ericcf: @michaelwehrley the convention is to prefix with `.` for class methods, and `#` for instance methods right?

michaelwehrley: arg!  You are right.  Amateur hour!

PR #263, opened 2015-10-14
Change Password Validations

michaelwehrley:
* Change password validations from `validate` to `before_validation`.
* Update specs to verify when password is blank, that an error is raised.
* Move duplicate Devise code into concern.
* Update spec syntax based on PR.

[#98987112]


app/models/user.rb
```diff
@@ -1,12 +1,6 @@
-require "devise"
-
 # A person with some authoritative role (a non-Participant).
 class User < ActiveRecord::Base
-  include ThinkFeelDoEngine::Concerns::ValidatePassword
-
-  devise :database_authenticatable,
-         :recoverable, :trackable, :validatable, :timeoutable,
-         timeout_in: 20.minutes
+  include ThinkFeelDoEngine::Concerns::Authentication
```
ericcf: I don't think this refactoring is warranted - if we ever want to make requirements different for Users vs Participants this will have to be undone.

ericcf: The validation could be shared, but I don't think Devise config should be.

michaelwehrley: Even though we haven't yet made a distinction in months if not years?

ericcf: Yes - I would like to be able to add NU netid authentication at some point, e.g.

michaelwehrley: :sound: good to me.  Thanks!

PR #264, opened 2015-10-14
Move Rake Dependencies Into Engine

michaelwehrley:
* Move notification rake file into engine
* Move Lesson mailer along with lesson views to engine
* Move lesson mailer html and text views
* Added named parameters to mailer

[#105006670]


ericcf:
:hotsprings:

PR #265, opened 2015-10-15
Add Lesson Notification Mailer

michaelwehrley:
* Add spec to verify notification mailer
* This spec was moved from host apps.

[#105006670]


eschlange:
:+1:

PR #266, opened 2015-10-15
Fix Lesson Notification From Bug

michaelwehrley:
* Use the from configuration to send emails, not application name.

[#105006670]


michaelwehrley:
I wasn't thinking! ARG!

spec/dummy/config/environments/test.rb
```diff
@@ -35,7 +35,7 @@
   # ActionMailer::Base.deliveries array.
   config.action_mailer.delivery_method = :test
   config.action_mailer.default_url_options =
-    { host: "stepped_care-no-reply@northwestern.edu" }
+    { host: "stepped_care.northwestern.edu" }
```
michaelwehrley: @ericcf FYI, I know this has no impact on the real application, but I needed to do this so things made more sense in testing.

ericcf: Yeah, the host probably shouldn't be an email address! I expect this was originally a case of copy and paste.

ericcf:
:nice: :black_joker:

PR #267, opened 2015-10-20
Msw 98987112 update password js

michaelwehrley:


spec/models/password_validator_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+        end
+      end
+    end
+
+    describe "when password passes model validations" do
+      it "calculates entroy" do
+        pv = PasswordValidator
+             .new(password: "abc123!Getloud!", password_token: token)
+
+        expect(pv.entropy_value).to eq 18.625
```
eschlange: Does this entropy value signify that this test's password value is significantly under the required entropy for our password system? If so, it seems that we may be setting the required entropy value too high.

eschlange:
Aside from my question, this looks clean. Plenty of tests. Feel free to merge.

spec/models/password_validator_spec.rb
michaelwehrley: This is actually just a test to make sure everything is working.

PR #270, opened 2015-10-22
Update Password Validator

michaelwehrley:
* Address bug when not able to find a `User` or `Participant` via token.

[#98987112]


eschlange:
:+1:

app/models/think_feel_do_engine/password_validator.rb
```diff
@@ -28,12 +28,9 @@ def calculate_entroy
...
     def find_or_intialize_person
-      if token
-        @person ||= Participant.find_by_reset_password_token(token) ||
-                    User.find_by_reset_password_token(token)
-      else
-        @person = Participant.new
-      end
+      @person ||= Participant.find_by_reset_password_token(token) ||
+                  User.find_by_reset_password_token(token) ||
+                  Participant.new
```
ericcf: @michaelwehrley I don't understand this code. What if it's a User trying to reset without a correct token? Then you're initializing a Participant.

michaelwehrley: @ericcf Exactly.  Just initializing them because there is no token, I am not creating them.  I want to use the model validations so all validations are the same with respect to the password.

michaelwehrley: but you indirectly make me think of a person updating their password if they are already logged in.  I will double check that.

ericcf: I guess I don't understand why you bother looking up a record by token. Wouldn't this work exactly the same way if you replaced the `find_or_initialize_person` method with

```ruby
@person ||= Participant.new
```
?

michaelwehrley: If a correct token exists, then I check the password against their email.

ericcf: Thanks. Definitely not obvious from the code! Probably worth adding a spec to test against email?

michaelwehrley: Way ahead of you buddy: https://github.com/cbitstech/think_feel_do_engine/blob/master/spec/models/think_feel_do_engine/concerns/validate_password_spec.rb#L80-L101

PR #271, opened 2015-10-27
Update Password Validator

michaelwehrley:
* On password reset page, update the action route
* Route is dynamic depending on whether person is user or participant.
* Update user and participant specs to look for dynamic strength

[#98987112]


app/assets/javascripts/think_feel_do_engine/devise/passwords/edit.js.erb
```diff
@@ -1,7 +1,7 @@
 <% url = ThinkFeelDoEngine::Engine.routes.url_helpers %>
 
 $(document).on('page:change', function() {
-  var password_path = '<%= url.participant_password_path %>';
+  var password_path = location.pathname.replace('/edit', '');
```
michaelwehrley: @ericcf This feels a bit like a hack, but I am not sure what would be a better option since a person isn't logged in at this time.

ericcf: What happens when you use `url.participant_password_path` for an unauthenticated user?

michaelwehrley: The jQuery code below can't find the correct `action` attribute for the form and doesn't validate the password for users dynamically.

ericcf: Yeah, wouldn't the action be `participant_password` for participants and `user_password` for users?

ericcf:
Ok, thanks for the updates! :100:

PR #272, opened 2015-10-28
Msw 100127766 revise stepping range

michaelwehrley:


ericcf:
I'm confused by the structure of `validate_data_reliability`. It looks like there's an `else` without an if...

michaelwehrley:
the else is part of the if block, but the `end` that you are thinking is going with the `if` is actually for the `do`: https://github.com/cbitstech/think_feel_do_engine/blob/msw_100127766_revise_stepping_range_2/app/models/phq_stepping.rb#L93-L95

ericcf:
:mind blown:

michaelwehrley:
Chris Lee blew your mind! :smile:

PR #273, opened 2015-10-28
Return nil when no event exists

michaelwehrley:
* If no `stop_event` exists, then `nil` needs to be returned.
* Add spec to verify that `nil` is returned.
* `nil` is returned based on the comments mentioned in code.

[Finished: #106743200]


spec/models/think_feel_do_engine/reports/video_session_spec.rb
```diff
@@ -48,6 +48,24 @@ module Reports
             )
           end
         end
+
+        context "the only following events are videoPlay" do
+          let(:event) { event_capture_events(:event_capture_events_179) }
+
+          it "returns nil when no 'stop event' exists" do
+            expect(EventCapture::Event
```
eschlange: Does this method chain represent a stop event? If so, should this logic exist as a method or scope within the event model?

michaelwehrley: interesting point!

PR #276, opened 2015-10-30
Msw 106947446 update froogaloop2

michaelwehrley:
* The first commit  is the most helpful in looking at code changes.
* The second commit is just formatting.  Thanks!

[Finished: #106947446]


eschlange:
@michaelwehrley I found this (http://stackoverflow.com/questions/21230170/failed-to-execute-postmessage-error-from-froogaloop-via-angularjs-directive) Did you try executing your code on the "load" event rather than "ready"?  According to this (http://stackoverflow.com/questions/8396407/jquery-what-are-differences-between-document-ready-and-window-load) it seems as though the load event may be further down the page lifecycle than "ready".

michaelwehrley:
@eschlange I did try that, and it still didn't work.  At some point I would love to figure why turbo links doesn't for for this function.  Thanks!

eschlange:
@michaelwehrley In that case, I'll merge. Since this setting is localized, I don't see a problem with this.

PR #277, opened 2015-11-02
Update Group Scope

michaelwehrley:
* Update group scope to correctly count login events.
* Reuse membership login count.

[Finished: #106168974]


eschlange:
Very clean~

PR #278, opened 2015-11-02
Add Correct Email Messaging References

michaelwehrley:
* Add spec to check for correctly referenced url.
* Change tool reference from 'MESSAGES' to tool title.
* Update fixtures to tests tool title with a space.

[Finished: #105498898]


app/mailers/think_feel_do_engine/message_notifications.rb
```diff
@@ -8,6 +8,12 @@ def new_for_coach(coach, group)
     end
 
     def new_for_participant(participant)
+      @context_name = participant
+                      .current_group
+                      .arm
+                      .bit_core_tools
+                      .find_by_title("MESSAGES")
```
ericcf: Won't it be very brittle to search by a hard-coded database value that could change?

ericcf: Maybe better to search by the `type` field?

db/migrate/20151102160119_add_slug_to_bit_core_tools.rb
```diff
@@ -0,0 +1,17 @@
+class AddSlugToBitCoreTools < ActiveRecord::Migration
+  def up
+    add_column :bit_core_tools, :slug, :string
```
ericcf: Won't it be more difficult to write logic that uses this field if it's allowed to be null?

app/mailers/think_feel_do_engine/message_notifications.rb
michaelwehrley: Good point.  I forgot about the type field.  Thanks!

db/migrate/20151102160119_add_slug_to_bit_core_tools.rb
ericcf: Sorry, just saw line 11!

michaelwehrley: I think there is a different solution that may be better.  I will update with your comment and resubmit.  Thanks!

michaelwehrley:
@ericcf The url wasn't even the problem.  I had to fix the url reference which I did, and then stupidly got caught up in the slug idea.  I already did test out on staging that `https://sunnyside-staging.cbits.northwestern.edu/navigator/contexts/Contact%20Moderator` works.  No problem.

ericcf:
Thanks!

PR #279, opened 2015-11-03
Add Scope for Ordering

michaelwehrley:
* Add scope to enabling find the most recent phq assessment

[Finished: #100127766]


app/models/phq_assessment.rb
```diff
@@ -11,6 +11,12 @@ class PhqAssessment < ActiveRecord::Base
   validates :release_date, uniqueness: { scope: :participant_id }
   validate :scores_valid
 
+  scope :when_updated, -> { order(updated_at: :desc) }
```
ericcf: Would you consider renaming to something more intuitive such as `latest_updated`?

michaelwehrley: that is perfecto.  :bow: thanks!

michaelwehrley:
@ericcf all set.

ericcf:
:cake:

PR #280, opened 2015-11-04
Add Background to Home Screen Thought Viz

michaelwehrley:
* Dynamically show harmful thoughts based condition.
* Via migration, update `type` to Think Tool.
* Add info notice if the number of harmful thoughts is less than 3.

[#100063566]


michaelwehrley:
Some conditions in the host app are necessary:
1. install migration
2. Update https://github.com/cbitstech/sunnyside/blob/master/app/models/presenters/content_module_presenter.rb#L4 to include `Tools::Think`

app/views/think_feel_do_engine/thoughts/distortion_viz.html.erb
```diff
@@ -1,6 +1,12 @@
-<div id="ThoughtVizContainer">
-  <div id="ThoughtVizTooltip"></div>
-</div>
+<h4>Harmful Thoughts</h4>
+<% harmful_pattern_ids = thoughts.harmful.map(&:pattern_id).compact %>
+<% if harmful_pattern_ids.uniq.select { |id|  harmful_pattern_ids.count(id) > 2 }.present? %>
```
ericcf: I'd like to see either a view spec for this, or a presenter with specs applied. What do you think about that?

michaelwehrley: I was thinking that a feature spec "should" have existed already for the js that displays the content and thus I didn't think it was necessary, but I will right a spec.

michaelwehrley: @ericcf That was good advice.  :enjoyable.  Thanks.

ericcf:
:tada:

PR #281, opened 2015-11-05
Update Scope to Find By Release Date

michaelwehrley:
* Rename scope
* Update scope to find the most recently updated
* Add specs to test scope
* Add fixtures, b/c dynamically creating phqs caused false errors.

[Finished: #107392650]


app/models/phq_assessment.rb
```diff
@@ -11,10 +11,10 @@ class PhqAssessment < ActiveRecord::Base
   validates :release_date, uniqueness: { scope: :participant_id }
   validate :scores_valid
 
-  scope :latest_updated, -> { order(updated_at: :desc) }
+  scope :recenty_released, -> { order(release_date: :desc).limit(1) }
```
ericcf: Why limit this to one record?

michaelwehrley: I figured, what is the point to grab all of them if I only need one.

ericcf: These ActiveRecord methods are lazily evaluated: it won't touch the database until you start iterating. It seems like the scope would be more flexible without the limit.

michaelwehrley: but won't it start iterating over the whole collection vs just one record?

ericcf: I'll leave that as an exercise for the reader.

ericcf:
:rotating_light:

app/models/phq_assessment.rb
michaelwehrley: :smile:

PR #282, opened 2015-11-05
Add Message Validation

michaelwehrley:
* Add validation to check if participant is messageable
* Add model specs to check if participant AND if arm implements messaging

[Finished: #107436990]


app/models/message.rb
```diff
@@ -30,11 +32,26 @@ def render_body
...
+      .current_group
+      .arm
+      .bit_core_tools
+      .find_by_type("Tools::Messages")
+  end
+
+  def participant_is_messageable
+    unless messaging_tool?
+      errors
+        .add(:base, "We're sorry, but this participant can't receive emails.")
```
michaelwehrley: @cbrenner04 can you confirm that this language is okay?  If not, can you provide the correct language. Thanks.

cbrenner04: I think "We're sorry, but this participant does not have access to the messaging tool." is more informative.

michaelwehrley: @cbrenner04 thanks you.

app/models/message.rb
```diff
@@ -30,11 +32,27 @@ def render_body
...
 
-  def populate_sent_at
-    self.sent_at = Time.new
-  end
-
   def create_delivered_messages
     recipient.received_messages.create(message_id: id)
   end
+
+  def messaging_tool?
```
ericcf: Why the `?` in the name. Usually the convention is for methods that return booleans to have a `?`.

michaelwehrley: bless it, after refactoring I removed the `false`.  :lame.

michaelwehrley: well because it returns truthy or falsey.

michaelwehrley: but I can fix that.

ericcf: All Ruby methods return truthy or falsy :-)

michaelwehrley: LOL, that is ridiculously truthy.

PR #283, opened 2015-11-20
Add Sanitize Method

michaelwehrley:
* This allows safer html to be parsed and displayed for editing
* Add feature spec to test

[Finished: #108576746]


app/views/think_feel_do_engine/bit_maker/slides/_fields.html.erb
```diff
@@ -30,7 +30,7 @@
   <div class="row">
     <div class="col-md-6">
       <div class="editable" contenteditable="true">
-        <%= slide.body %>
+        <%= sanitize(slide.body) %>
```
ericcf: Why not just call `render_body`?

michaelwehrley: Good question. I had to check it out and I could update `render_body` in https://github.com/cbitstech/bit_core/blob/d2f0fe79463a11c7f378df79a4ce2ea3ca2d37cf/app/models/bit_core/slide.rb to use `sanitize`.  However, the way we currently have it, the html `<h1>Hello world</h1>` content gets converted to `&lt;h1&gt;Hello world&lt;/h1&gt;`.  Thoughts on the best option?

ericcf:
:100:

PR #285, opened 2015-12-08
Reduce password entropy to allow for simpler passwords.

eschlange:
[#108985146]


ericcf:
Can you update the specs here with a password that has entropy between 10 and 18: https://github.com/cbitstech/think_feel_do_engine/blob/master/spec/models/think_feel_do_engine/concerns/validate_password_spec.rb ?

eschlange:
All set, I added a spec that validates the success of a password creation with an entropy value of 11.

eschlange:
@cbrenner04 @ericcf I attempted to programmatically trigger a password reset and directly navigate to the page with this code:

```
def navigate_to_change_password_page
    user = User.find(users(:admin1).id)
    raw_token, user.reset_password_token =
      Devise.token_generator.generate(User, :reset_password_token)
    user.save

    visit "/users/password/edit?reset_password_token=[#{raw_token}]"

    expect(page).to have_text "Change your password"
  end
```

It succeeds intermittently on each of the tests, and also it runs slightly slower than resetting the password by navigating to the password reset page. I'm going to leave this as-is, since this is a feature spec, it seems like this should be acceptable as it stands.

ericcf:
:thought_balloon:

PR #287, opened 2015-12-14
Update Spec Counts

michaelwehrley:
* Improve brittleness by checking event counts are greater than

[#98199340]


spec/models/think_feel_do_engine/reports/event_spec.rb
```diff
@@ -22,7 +22,7 @@ def data
           it "returns accurate summaries" do
             event = event_capture_events(:event_capture_events_010)
 
-            expect(data.count).to eq 153
+            expect(data.count).to be >= 153
```
cbrenner04: @michaelwehrley what if fixtures are removed? I would think we would want the comparison to be low enough where we can add or subtract data a reasonable amount and this would still pass. This is why I suggested comparing to 100

michaelwehrley: @cbrenner04 sounds good.

PR #288, opened 2015-12-14
Update existing participant password feature spec.

eschlange:
* Broke existing feature spec down into seperate tests to remove race condition error.

[#108985146]


eschlange:
Merging in. Changes identical to User spec for same ticket ID.

PR #290, opened 2015-12-18
Create Learn Tool When Necessary

michaelwehrley:
* When no learn tool exists, create it when creating a learning slide
* Update specs to create lesson.

[#110321804]


app/controllers/think_feel_do_engine/lessons_controller.rb
```diff
@@ -102,6 +105,14 @@ def lesson_params
       params.require(:lesson).permit(:title, :position) if params[:lesson]
     end
 
+    def set_lesson_tool
+      @lesson_tool ||= @arm
+                       .bit_core_tools
+                       .find_or_create_by!(
```
michaelwehrley: @ericcf The alternative is to make sure that the database has a lesson tool.  Not sure which way is better.

michaelwehrley: @ericcf This is also occurring here and could use a similar fix: https://github.com/cbitstech/think_feel_do_engine/blob/master/app/controllers/think_feel_do_engine/bit_maker/slideshows_controller.rb#L32

PR #292, opened 2015-12-28
Update Default Display Name

michaelwehrley:
* Participant's display name returns `nil` if not set.

[#108473648]


app/models/participant.rb
```diff
@@ -91,11 +91,7 @@ def self.inactive
   }
 
   def display_name
```
ericcf: Why does this method need to be overridden now?

michaelwehrley: Good point (or question I guess).  I forgot that AR has that an attribute as a default method.

ericcf:
Lines of code deleted = win! :neckbeard:

michaelwehrley:
heck yeah! :sagittarius:

PR #293, opened 2015-12-30
Catch and Redirct Unrecognized Routes

michaelwehrley:
* Catches and redirects the user when a route is not defined.

[#106849832]


config/routes.rb
```diff
@@ -97,6 +97,12 @@
 
   get "password_entropy_bits", to: "password_entropy_bits#show", as: "password_entropy_bits"
 
+  # catch all for unrecognized routes
+  if Rails.env == "production"
+    get "*unknown", to: redirect("/404")
```
ericcf: Any reason not to do this on staging as well?

michaelwehrley: Nice catch!

PR #296, opened 2016-01-05
Optimize Wildcard Route Catching

michaelwehrley:
* Move routing below quick sign in.
* Remove conditions.

[#110664746]
[#111014166]


michaelwehrley:
Silly oversight.

PR #298, opened 2016-01-06
Catch UrlGenerationError

michaelwehrley:
* Rescue url generation error and set to "/"
* Update routes and spec dummy routes for jasmine app

[#111014166]


ericcf:
Looks great, thanks!

PR #299, opened 2016-01-14
Update Failing Spec

michaelwehrley:
* Password field should not be validated if it hasn't changed
* Add spec to check if password field isn't being updated

[#111701452]


eschlange:
Lookin' good. I'll merge when the build finishes.

michaelwehrley:
@eschlange  Thanks!  I was a little concerned about this updated, so thanks for looking it over.

PR #303, opened 2016-01-21
Redirect When Record Not Found

michaelwehrley:
* Add `rescue` when records aren't found
* Redirect logged in user if records aren't found
* Add spec coverage
* Improve specs via instance doubles and nesting

[#112063427]


eschlange:
Solid controller changes and tests.

PR #304, opened 2016-01-22
bump devise version for security patch

ericcf:


ericcf:
I'm going ahead and merging because specs passed.

PR #307, opened 2016-02-02
Reduce Route Footprint

michaelwehrley:
* remove unnecessary routes
* Update cancancan requirement

[#112597249]


ericcf:
:thought_balloon: good

PR #310, opened 2016-02-08
Positive Experience

michaelwehrley:
* Add `Experience` class
* Add view to enter a positive experience
* Add view to display all positive experiences
* Add content provider to display experience form
* Add model, view, and provider spec

[#100210978]
[#110761812]
[#110761932]


ericcf:
Nice! Very thorough. :horse:

PR #311, opened 2016-02-09
Allow for Pulic View

michaelwehrley:
* skip authentication for public slideshow
* Rescue if slideshow isn't found
* Add controller spec for authentication and redirect

[Finishes: #113284567]


app/controllers/think_feel_do_engine/application_controller.rb
```diff
@@ -65,6 +65,17 @@ def verified_request?
 
     private
 
+    def authenticate_person!
```
ericcf: I think there should be unit tests for this method you added.

michaelwehrley: good call.

PR #314, opened 2016-02-12
Rescue if Token is Invalid

michaelwehrley:
* Add rescue if token is invalid
* Add concern with rescue statement
* Include concern in controllers
* Add controller spec to test for authentication bug

[#113630033]


michaelwehrley:
Just a side note, b/c this code has to be in the host apps (if they have additional controllers) and in multiple engines, I thought a ruby gem might be a good idea.  I worked on this for bit to see how it would go: https://github.com/michaelwehrley/rescue

ericcf:
Do you want someone to look at this?

michaelwehrley:
Not the gem at this point, but I will submit the PR in a moment.  Thanks!

app/models/think_feel_do_engine/concerns/invalid_auth_token.rb
```diff
@@ -0,0 +1,19 @@
+module ThinkFeelDoEngine
```
ericcf: What was the reason for putting this code in `models`? It seems more related to controllers.

spec/controllers/think_feel_do_engine/participants/task_status_controller_spec.rb
```diff
@@ -41,6 +41,23 @@ module Participants
 
             it { expect(response.status).to eq 400 }
           end
+
+          describe "when authorization token has expired" do
```
ericcf: What about writing a shared example to test the redirect logic? Then you could easily test any controllers that inherited the module.

app/models/think_feel_do_engine/concerns/invalid_auth_token.rb
michaelwehrley: Good point; it should be in a controllers concerns folder.

spec/controllers/think_feel_do_engine/participants/task_status_controller_spec.rb
michaelwehrley: Great idea!

michaelwehrley: I never knew how helpful `shared_examples` were.  Great recommendation!

spec/controllers/think_feel_do_engine/participants/task_status_controller_spec.rb
```diff
@@ -42,6 +42,10 @@ module Participants
             it { expect(response.status).to eq 400 }
           end
         end
+
+        it_behaves_like "an untrusted action when auth token has expired" do
```
ericcf: Nice. Maybe add this shared example to a passwords controller spec and sessions controller spec?

ericcf:
@michaelwehrley looks good, thanks!

PR #317, opened 2016-02-25
Memberbership Date Methods Include Timezone

michaelwehrley:
* In order to have scopes act correctly, timezones must be present
* Have `week_start_day` and `week_end_day` include timezones

[#113961399]


app/models/membership.rb
```diff
@@ -196,4 +196,8 @@ def not_ending_in_the_past
 
     errors.add :end_date, "must not be in the past"
   end
+
+  def start_date_with_zone
+    start_date.in_time_zone
```
ericcf: What time zone is `start_date` in originally?

michaelwehrley: It is in instance of `Date` and when comparing in scopes, it needs to have more precision.  For instance
```
SELECT COUNT(*) FROM "social_networking_on_the_mind_statements" WHERE "social_networking_on_the_mind_statements"."participant_id" = 64 AND ("social_networking_on_the_mind_statements"."created_at" >= '2016-02-05 06:00:00.000000') AND ("social_networking_on_the_mind_statements"."created_at" < '2016-02-12 06:00:00.000000')
```
vs
```
SELECT COUNT(*) FROM "social_networking_on_the_mind_statements" WHERE "social_networking_on_the_mind_statements"."participant_id" = 74 AND ("social_networking_on_the_mind_statements"."created_at" >= '2016-02-05') AND ("social_networking_on_the_mind_statements"."created_at" < '2016-02-12')
```

ericcf: Ok, so you should get the same result with `.to_time`: http://api.rubyonrails.org/classes/Date.html#method-i-to_time

Rails automatically assigns the local timezone.

michaelwehrley: @ericcf That is what I needed!  Thanks!

spec/models/membership_spec.rb
```diff
@@ -150,14 +150,14 @@
     describe "#week_start_day" do
       it "returns the first day of the week" do
         expect(membership1.week_start_day(week_number))
-          .to eq Date.today + 7.days
+          .to eq((Date.today + 7.days).to_time)
```
ericcf: Here you're basically just rewriting the method logic in your test. In these cases my approach is usually to test with known input and output values instead so that the method remains a black box.

michaelwehrley: yeah, that it is a good point.

michaelwehrley:
@ericcf this is ready for review again.  Thanks!

ericcf:
:toilet:

PR #322, opened 2016-02-29
Add SlideshowProvider

michaelwehrley:
* No longer have `LessonModule` check first provider to determine if it has a `SlideshowProvider`
* Have `LessonModule` check for a the exist of a `SlideshowProvider` via `find_by`
* Add spec to check for creation of `SlideshowProvider`

[Finishes: #114693699]


app/models/content_modules/lesson_module.rb
```diff
@@ -1,6 +1,7 @@
 module ContentModules
   # Container for didactic content.
   class LessonModule < BitCore::ContentModule
+    CONTENT_PROVIDER_TYPE = "BitCore::ContentProviders::SlideshowProvider"
```
ericcf: Would it be clearer to name this `SLIDESHOW_PROVIDER_TYPE` or similar?

spec/models/content_modules/lesson_module_spec.rb
```diff
@@ -34,5 +34,29 @@ module ContentModules
         end
       end
     end
+
+    describe ".build_slide" do
+      describe "when the first content provider isn't of type `SlideshowProvider`" do
+        it "will add a slideshow content provider" do
+          BitCore::ContentProvider.create!(
```
ericcf: Any reason not to `ContentProviders::LearnLessonsIndexProvider.create!(...` instead?

michaelwehrley: nope, seems helpful

app/models/content_modules/lesson_module.rb
michaelwehrley: i like. thanks.

ericcf:
:shipit:

PR #324, opened 2016-03-04
Add markdown rendering to site messaging.

eschlange:
* Added Redcarpet markdown renderer to site message mailer.
* Added html escaping to site message mail template.
* Added spec to ensure rendering was a success.

[#114009049]


ericcf:
Looks good. :shipit: when Travis says.

PR #325, opened 2016-03-08
Add is_admin to event report.

eschlange:
- added is admin detail to event all query.
- updated participant fixture to include is_admin field.
- updated event spec to expect new field.

[#115279831]


michaelwehrley:
Excellent work! :rocket:

PR #326, opened 2016-03-09
Add column header for is_admin field.

eschlange:
[#115279831]


eschlange:
Very small change, merging.

PR #327, opened 2016-03-10
Add handling for missing slides.

eschlange:
* Bypassing nil exceptions from deleted slides.
* Added spec for new logic.

[#114775707]


app/models/think_feel_do_engine/reports/lesson_slide_view.rb
```diff
@@ -87,7 +87,7 @@ def self.find_slide_by_url(slides, url)
           slide_position = url[/\d+$/].to_i
         end
 
-        slides.find_by_position(slide_position)
+        slides.find_by_position(slide_position) unless slides.nil?
```
michaelwehrley: Do we want `nil?` here or `empty?`?  Isn't `slides` a relation?

michaelwehrley:
@eschlange what happens at https://github.com/cbitstech/think_feel_do_engine/blob/ems-114775707/app/models/think_feel_do_engine/reports/lesson_slide_view.rb#L42 if a lesson isn't found?  Should this be a `try` statement?

app/models/think_feel_do_engine/reports/lesson_slide_view.rb
eschlange: Find_by returns a record (or nil if nothing is found.) http://api.rubyonrails.org/classes/ActiveRecord/FinderMethods.html#method-i-find_by

eschlange:
@michaelwehrley The spec that I wrote for this logic tests that scenario.  In the case where no lesson is found, it doesn't add to the report. Missing lessons are what was causing the issue to begin with.

app/models/think_feel_do_engine/reports/lesson_slide_view.rb
```diff
@@ -39,17 +42,20 @@ def self.all_slide_interactions
...
-
-            {
-              participant_id: participant.study_id,
-              lesson_id: lesson_id,
-              slide_id: slide.try(:id),
-              slide_title: slide.try(:title),
-              slide_selected_at: e.emitted_at.try(:iso8601),
-              slide_exited_at: next_event_at(e).try(:iso8601)
-            }
+            begin
```
michaelwehrley: :santa:

michaelwehrley:
I think that is more precise.  Thanks!

michaelwehrley:
:guitar: :shipit: when passes!

PR #328, opened 2016-03-11
Redirect new site message creation for no participants.

eschlange:
* Updated controller new logic.
* Added spec.

[#114007863]


app/controllers/think_feel_do_engine/coach/site_messages_controller.rb
```diff
@@ -20,6 +20,9 @@ def show
 
       def new
         @participants = current_user.participants_for_group(@group)
+        redirect_to(coach_group_site_messages_path(@group),
+                    notice: "A group must have participants in order "\
```
ericcf: Should this be a `notice` or an `alert`?

eschlange: Ah yes, an alert makes more sense. Updating now.

spec/controllers/think_feel_do_engine/coach/site_messages_controller_spec.rb
```diff
@@ -46,12 +46,21 @@ module Coach
           end
 
           it "responds with the group's participants but scoped to the current user" do
```
michaelwehrley: Should this test be titled something along the lines of that 'it responds with new template'?  This test actually has no idea if the participant is scoped to the `current_user` given the test setup.

eschlange: I agree, I'll update the description of this test.

michaelwehrley:
:+1:

PR #329, opened 2016-03-15
Week-Based Content Modules

michaelwehrley:
* Allow for content modules to display within panels based on week criteria
* Add scopes and scope specs for time-based module panels
* Fix `TaskHelper`/`TaskStatusLink` method bug
* Add `TaskStatusLink` spec for when a task status has been completed
* Add feature spec to check correct module layout

[#114007819]


app/models/tools/practice.rb
```diff
@@ -0,0 +1,5 @@
+module Tools
+  # Displays other tools in the form of content providers
+  class Practice < BitCore::Tool
```
ericcf: Is there special behavior based on this tool class?

michaelwehrley: Hmmm, are you suggesting that I just use an already existing tool and change the title?

ericcf: Not suggesting anything yet, just asking :-)

michaelwehrley: I am actually not sure if it will have any special behaviors in the future.  I needed a class defined for the fixtures to load and I wanted to keep things consistent with names of tools and the title in the toolbar.  The functionality is a tool that lists a bunch of other `ContentProviders` that participants need to access - kind of an index page per se.  These providers display on one side `AvailableContentMondules` that are assigned for the current week and on the other side `AvailableContentMondules` that were assigned in the past.  Let me know if that helps :smile_cat:

ericcf: Sounds like a case of YAGNI! I understand what you're saying, but it sounds like the functionality is accomplished with modules and providers. I'd say just make it a plain `BitCore::Tool` for now and change it later if needed.

michaelwehrley:
merging... :smile:

PR #331, opened 2016-03-16
Add Raven reporting and handle context issues with navigator.

eschlange:
* Bypassing URLGeneration error on malformed parameter set.
* Add spec to ensure redirect when module is not specified.

[#115355529]


app/controllers/think_feel_do_engine/navigator_controller.rb
```diff
@@ -106,5 +115,14 @@ def home_tool
         .arm
         .bit_core_tools.find_by_type("Tools::Home")
     end
+
+    def next_content_not_found
```
michaelwehrley: I don't know if you need Raven here.  We define it and set it in all host app application controllers: https://github.com/cbitstech/think_feel_do_so/blob/master/app/controllers/application_controller.rb#L16-L23

eschlange: I added it to give additional context with the parameters passed into the controller. I had trouble understanding how the exception occurred since the parameters are listed in the Raven exception. If this isn't going to help, then I'll remove it.

spec/controllers/think_feel_do_engine/navigator_controller_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+
+          expect(response).to redirect_to("/")
+          expect(flash[:alert]).to be_present
+        end
+      end
+    end
+  end
+end
+
+
```
michaelwehrley: rubocop :cop: missed this

app/controllers/think_feel_do_engine/navigator_controller.rb
michaelwehrley: I think if you didn't know enough from the context, then adding params seems reasonable :sagittarius:

michaelwehrley:
:+1: good to go!  Nice works and spec.  Thanks!

app/controllers/think_feel_do_engine/navigator_controller.rb
eschlange: Also, this is not due to an exception being raised. I added this to indicate that there is an issue with content module data.

eschlange:
Thanks for the comments! I'll merge it in after the updates based on your comments have passed CI.

PR #332, opened 2016-03-16
Add Completion Method

michaelwehrley:
* Add methods to determine if previous task status has been completed
* Add additional specs for new public methods
* Add specs for membership methods
* Remove unnecessary method and specs  `available_for_learning`

[#114008595]


app/models/task_status.rb
```diff
@@ -67,6 +56,22 @@ class TaskStatus < ActiveRecord::Base
...
+    available_for_learning_on <= Date.today
+  end
+
+  def completed?
+    completed_at.present?
+  end
+
+  def previous_completed?
+    return true if is_first?
+    self
```
eschlange: What is the purpose of invoking this method off of the model's class?

michaelwehrley: Sorry, but I am not sure I understand the question.  I am running a query here and I need to do that off the `TaskStatus`.  Are you saying to instead write:
```ruby
  def previous_completed?
    return true if is_first?
    TaskStatus
      .find_by_id(membership_task_status_ids[index_for_membership - 1])
      .completed?
  end
```

eschlange: Ah, understood. I haven't seen that notation used for this purpose.

eschlange:
Looks good, nice refactoring and specs as well.

PR #335, opened 2016-03-17
Msw update times spec

michaelwehrley:
This fixed the membership bug; so a small improvement.  However, with the seed (45362) provided in this travis run, the activity feature spec below still fails.  https://travis-ci.org/cbitstech/think_feel_do_engine/builds/116630760


eschlange:
:+1: Nice job on the refactoring.

PR #340, opened 2016-03-21
Handle record not found issues for showing content modules.

eschlange:
* Update to controller logic.
* Added spec for new logic.
* Handles missing task when updating task_status.
* Add spec to check for handling of exception.

[#115355465]
[#113190343]


spec/controllers/think_feel_do_engine/participants/task_status_controller_spec.rb
```diff
@@ -42,6 +42,17 @@ module Participants
             it { expect(response.status).to eq 400 }
           end
 
+          context "that specifies a non-existant task" do
+            before do
+              allow(task_status)
+                .to receive(:mark_complete)
+                .and_raise(ActiveRecord::InvalidForeignKey.new("invalid key"))
```
michaelwehrley: Can we just raise `ActiveRecord::InvalidForeignKey` or do we need to create a new one here?

michaelwehrley:
:+1: :shipit:

spec/controllers/think_feel_do_engine/participants/task_status_controller_spec.rb
michaelwehrley: looks like you can't.  I don't quite understand why you can't yet.

PR #341, opened 2016-03-21
Add check to ensure intro slideshow has slides.

eschlange:
[#113404069]


eschlange:
Merging, errors due to external spec issues.

PR #342, opened 2016-03-22
Ems 114719255 - Abstract out position numbers from lesson view.

eschlange:
* Update sorting icon to be less confusing.
* Removed position number from view and edit pages.
* Removed spec logic expecting position numbers.
* Removed feature spec that had a db race condition issue.

[#114719255] [#116130777]


eschlange:
Merging, mostly code removal.

PR #343, opened 2016-03-22
Fix MediaAccessEvent date calculations.

eschlange:
* Now uses group specific to dashboard page.
* Updated specs to expect a group parameter.

[#115859907]


ericcf:
:cool: :sunglasses:

PR #345, opened 2016-04-01
Remove ability to switch reply recipient.

eschlange:
- Setting recipient on reply form if given.
- Added controller spec to ensure recipient is set.

[#116478101]


michaelwehrley:
makes sense to me. :shipit:

app/controllers/think_feel_do_engine/coach/messages_controller.rb
```diff
@@ -23,6 +23,8 @@ def new
         authorize! :new, Message
 
         @new_message = current_user.build_sent_message
+        @reply_recipient =
+          Participant.find(params[:recipient_id]) if params[:recipient_id]
```
michaelwehrley: did you also want to scope this? `current_user.participants.find(params[:recipient_id]) if params[:recipient_id])

michaelwehrley: You could also change: `authorize! :new, @new_message` and place `@new_message = current_user.build_sent_message` before the authorize call.

eschlange:
@michaelwehrley Thanks for the comments. I updated the controller using your two recommendations (I scoped the participant list and placed the auth code after the message object is build.) Lastly, I updated the specs to accommodate the changes.

Please let me know if everything looks good.

spec/controllers/think_feel_do_engine/coach/messages_controller_spec.rb
```diff
@@ -64,7 +74,7 @@ module Coach
 
           it "should set the recipient participant if given" do
             allow(Group).to receive(:find) { group }
-            expect(Participant).to receive(:find)
+            allow(auth_user).to receive_message_chain(:participants, :find) { [] }
```
michaelwehrley: This doesn't really matter, but I think this returns an participant or `nil`, not an array.

michaelwehrley:
Looks good to me.  My comment is minor :sheep:

eschlange:
@michaelwehrley Updated the spec to return a typed double, rather than an array. Mind if I merge it?

michaelwehrley:
Good work!

eschlange:
Thanks, Mike!

PR #346, opened 2016-04-04
Msw 114008595 release the lesson kraken

michaelwehrley:


ericcf:
Seems reasonable to me.

PR #347, opened 2016-04-06
Update lesson view report to include default URLS.

eschlange:
* Report now takes default lesson pages into account.
* Added spec that checks for default lesson URL.

[#116928465]


ericcf:
Looks good to me! I find the reporting code to be very painful to work on, so thanks.

PR #348, opened 2016-04-06
Handle missing groups for groups controller.

eschlange:
- Added RecordNotFound exception handling.
- Added a spec to ensure exception is caught and handled.

[#116918903]


michaelwehrley:
looking good. :heart_eyes:  Thanks for fixing that and adding a spec!

PR #350, opened 2016-04-07
Expand Record Not Found Method

michaelwehrley:
* Expand record not found method to determine if a record can or can't be found
* Add test to verify redirect

[Finishes: #117180921]


ericcf:
🍨

PR #352, opened 2016-04-08
Refactor patient dashboard logic, add event index.

eschlange:
- Updated lesson_module logic to include content modules.
- Cached a set of content modules in the lesson viewing logic.
- Added index to event table, it is fairly large on staging and prod.

[#117182071]


app/models/think_feel_do_engine/reports/lesson_module.rb
```diff
@@ -3,20 +3,20 @@ module Reports
...
+      # rubocop:disable Metrics/MethodLength
       def self.included(klass)
         class << klass
           # Returns a hash mapping lesson entry (first slide) path to Lesson
           # Module id.
           def lesson_entries_map
-            lessons = ContentModules::LessonModule.all.map do |m|
+            lessons = ContentModules::LessonModule
+                      .includes(:content_providers).map do |m|
               first_provider_id = m.provider(1).try(:id)
```
michaelwehrley: Just a note: if this is `nil` I think an exception will be raised when the path helper tries to build the `navigator_location_path`

app/models/think_feel_do_engine/reports/lesson_module.rb
```diff
@@ -3,20 +3,20 @@ module Reports
...
-
+      # Disabling method length due to nested methods.
+      # rubocop:disable Metrics/MethodLength
       def self.included(klass)
         class << klass
           # Returns a hash mapping lesson entry (first slide) path to Lesson
           # Module id.
           def lesson_entries_map
-            lessons = ContentModules::LessonModule.all.map do |m|
+            lessons = ContentModules::LessonModule
```
michaelwehrley: Does this need to be scoped to the `ContentModules::LessonModule` of an arm? or is it fine if the report shows ALL `ContentModules::LessonModule`?

app/models/think_feel_do_engine/reports/lesson_module.rb
eschlange: If what is nil?

michaelwehrley:
:+1: :eyes: :hotsprings:

PR #354, opened 2016-04-12
Ems 113319087 - Add static binary swf file for jplayer.

eschlange:
* Added binary file.
* Updated jplayer view to specify new file location.
* Update dummy config to include new SWF asset. 
[#113319087]


app/views/think_feel_do_engine/shared/_jplayer.html.erb
```diff
@@ -38,7 +38,7 @@
         mp3: "<%= audio_url %>"
       });
     },
-    swfPath: "http://www.jplayer.org/2.9.2/js",
+    swfPath: "<%= asset_path('jquery.jplayer.swf') %>",
```
michaelwehrley: nice improvement.

michaelwehrley:
Thanks for the example of how to add this to the asset pipeline! :two_men_holding_hands:

PR #355, opened 2016-04-13
Removed can can helper.

eschlange:
* Now utilizes find method to determine group.

[#116918903]


eschlange:
Merging, very small change.

PR #357, opened 2016-04-14
Add Skip Links

michaelwehrley:
* Update participant experience by adding skip buttons
* This allows for a better flow from learning to practice
* Unify where text 'Skip' is defined

[#117549987]


ericcf:
🐇

PR #358, opened 2016-04-14
Add logic to allow for trailing slash in lesson entry urls.

eschlange:
* Add fixture to test logic.
* Updated specs to expect new lesson number.
* Update lesson_view logic.

[#117546565]


app/models/think_feel_do_engine/reports/lesson_viewing.rb
```diff
@@ -28,13 +28,17 @@ def self.all(*ids)
             .to_a.select do |e|
               lessons.keys.include?(e.current_url.gsub(URL_ROOT_RE, "")) ||
               lessons_default
-              .keys.include?(e.current_url.gsub(URL_ROOT_RE, ""))
+              .keys.include?(e.current_url.gsub(URL_ROOT_RE, "")[0...-1]) ||
+              lessons_default
```
michaelwehrley: @eschlange is this what you want?  It seems to be the same line written twice.

eschlange: Good catch!

michaelwehrley:
:eyeglasses: good to :meat_on_bone:  :+1:

PR #359, opened 2016-04-14
Add Confirm When Incomplete

michaelwehrley:
* Add Confirm message when feelings are left unanswered
* Add feature spec to test submission.

[#117547829]


michaelwehrley:
In the [past](https://github.com/cbitstech/think_feel_do_engine/commit/b7afee27b18dd29e22b90ae313911bddf66603f1) a race condition of some sort caused this test to be removed.  It may be best regardless to move this spec to **Marigold**.

ericcf:
🐩

PR #361, opened 2016-04-20
Update patient dash summary title for lessons read.

eschlange:
[#117703553]


eschlange:
Merging, small text change.

PR #363, opened 2016-04-26
Limiting activity title input field maxlength.

eschlange:
[#117933985]


eschlange:
Merging, small template changes.

PR #365, opened 2016-04-26
Coach Message Partial

michaelwehrley:
* Move coach assignment message into partial
* Add spec to make sure partial renders

[#118299053]


spec/features/participant/messages/coach_assignment_spec.rb
```diff
@@ -0,0 +1,40 @@
+require "rails_helper"
```
ericcf: Is there a reason you opted for a feature spec over a view spec?

michaelwehrley: That may be more reasonable.

spec/views/think_feel_do_engine/messages/index.html_spec.rb
```diff
@@ -0,0 +1,34 @@
+require "rails_helper"
+
+RSpec.describe "think_feel_do_engine/messages/index", type: :view do
+  let(:message) do
+    "You will be able to use this tool once you have been assigned a coach"
+  end
+  let(:received_messages) do
+    double("received_messages", unread: [], empty?: true, count: 0, each: nil)
+  end
+  let(:sent_messages) { double("sent_messages") }
```
ericcf: What's this used for?

spec/views/think_feel_do_engine/messages/show.html.erb_spec.rb
```diff
@@ -24,5 +24,27 @@
...
+
+    describe "when participant doesn't have an assigned coach" do
+      let(:message) do
+        "You will be able to use this tool once you have been assigned a coach"
+      end
+      let(:participant) do
+        instance_double(
+          Participant,
+          in_study?: true,
+          coach_assignment: nil)
```
ericcf: When you don't care about the return value, you can use `.as_null_object` on doubles as a shortcut.

michaelwehrley: Not sure I follow.  I was looking at [this](http://www.relishapp.com/rspec/rspec-mocks/v/3-4/docs/basics/null-object-doubles), but in this example, how would that help?

spec/views/think_feel_do_engine/messages/index.html_spec.rb
michaelwehrley: good catch.  Old code...I ❤️ view specs :sarcastic

spec/views/think_feel_do_engine/messages/show.html.erb_spec.rb
ericcf: Sorry you're right: you need a falsy value!

michaelwehrley:
@ericcf if [one set](https://travis-ci.org/cbitstech/think_feel_do_engine/builds/125938742) passes is that sufficient?

ericcf:
@michaelwehrley I guess, but I would argue that we convert spec/features/participant/activities/content_providers/important_activity_checklist_spec.rb into unit tests if it's that flaky.

PR #366, opened 2016-04-26
Update participant dashboard visualizations with group specifics.

eschlange:
* Added group expectation to controller methods.
* Updated routes.
* Updated patient vis helper to include group specification.
* Updated views that use the path helpers.
* Added specs to check for error when accessed by withdrawn participants.

[#118113669]


app/controllers/think_feel_do_engine/coach/participant_activities_visualizations_controller.rb
```diff
@@ -12,6 +12,7 @@ class ParticipantActivitiesVisualizationsController < ApplicationController
 
       def show
         @participant = Participant.find(params[:participant_id])
+        @group = Group.find(params[:group_id])
```
michaelwehrley: I think we should add a `rescue` here if `Group` can't be found.  Does that make sense?

app/controllers/think_feel_do_engine/coach/participant_thoughts_visualizations_controller.rb
```diff
@@ -8,6 +8,7 @@ class ParticipantThoughtsVisualizationsController < ApplicationController
 
       def show
         @participant = Participant.find(params[:participant_id])
+        @group = Group.find(params[:group_id])
```
michaelwehrley: And here.

app/helpers/think_feel_do_engine/coach/patient_dashboard_helper.rb
```diff
@@ -11,7 +11,7 @@ def breadcrumbs
         return unless VISUALIZATION_CONTROLLERS.include?(controller_name)
 
         dashboard_path = coach_group_patient_dashboard_path(
-          @participant.active_group,
+          @group,
```
michaelwehrley: Would it make sense to maybe actually find the group off `@participant.groups.find(params[:group_id])`

app/controllers/think_feel_do_engine/coach/participant_activities_visualizations_controller.rb
michaelwehrley: After looking this over, I think we should find group based off `@participant.groups.find(params[:group_id])` and include a `rescue` in both this and `participant_thoughts_visualizations_controller`.

eschlange: I'll add a catch-all for participant and group.

app/helpers/think_feel_do_engine/coach/patient_dashboard_helper.rb
eschlange: No, a withdrawn participant lists the group in that scope.

michaelwehrley:
♐ Good work!

PR #369, opened 2016-05-02
Remove feature spec and replace with view and model specs.

eschlange:
[#118350379]


spec/views/think_feel_do_engine/activities/_previously_planned.html.erb_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+    it "displays nothing if activity has no start time." do
+      render partial: "think_feel_do_engine/activities/previously_planned",
+             locals: {
+               activity: activity_no_start_time
+             }
+
+      expect(rendered).to be_empty
+    end
+
+    it "displays nothing if activity has no start time." do
```
michaelwehrley: I think you mean "end time"

michaelwehrley:
Thank you!

PR #371, opened 2016-05-02
Ecf 118044731 twilio callback

ericcf:
This will allow host applications to configure project specific callback URLs for tracking success/failure of SMS messages via Twilio.


app/models/message_sms_notification.rb
```diff
@@ -25,16 +25,26 @@ def environment
     Rails.env
   end
 
+  def status_callback_url
+    config.try(:status_callback_url)
```
michaelwehrley: I take it Twillio has urls that we are using outside our application to display statuses?  Otherwise, couldn't we make/use internal pages and routes within our host applications without setting the url in the config folder?

michaelwehrley:
Looks good! 💯

app/models/message_sms_notification.rb
ericcf: Not exactly. Twilio will POST back to a URL we provide after it attempts to send an SMS asynchronously. This way we can keep a record of message statuses by project in a central location.

PR #373, opened 2016-05-03
Add datatable config to the PHQ9 table for date sorting.

eschlange:
[#112240763]


app/views/think_feel_do_engine/coach/patient_dashboards/tables/_phq9_data.html.erb
```diff
@@ -42,6 +44,23 @@
 
 <script type="text/javascript">
   $(function() {
-    return $('.generic-data-table').dataTable();
+    return $('#phq_assessments').dataTable({
```
michaelwehrley: makes more sense.

michaelwehrley:
looks good.  I would just try to test it locally to make sure the sorting works and the jquery dataTable functions run without errors. 👍

PR #375, opened 2016-05-09
Reorder Skip & Next Buttons

michaelwehrley:
* Move 'Next' buttons to the right of the 'Skip' buttons

[Finishes: #118856771]


eschlange:
🆗

PR #376, opened 2016-05-09
Ems 118962341 - Update task status completion.

eschlange:
* Move task status completion logic into navigator controller.
* Add specs for navigator controller changes.
* Remove Engagement model.
* Remove all associated engagement creation and update logic.
* Remove task status controller which is no longer used.
* Remove related specs.
* Update existing migration that directly reference Engagement model.

[#118962341]


michaelwehrley:
Awesome updates!  I really like the improvements. 👍  🍮

PR #377, opened 2016-05-10
Add Root Page Link

michaelwehrley:
* Add link in views that will redirect participants to home page.
* Add view specs to check if link exists

[#118857381]


spec/support/view_spec_helpers.rb
```diff
@@ -0,0 +1,6 @@
+RSpec.shared_examples "a practice 'index' page" do
```
ericcf: I think it's probably overkill to add this helper and test that `link_to` works, but if you feel it's necessary...

eschlange:
Looks fine to me.

spec/support/view_spec_helpers.rb
michaelwehrley: You are probably right.

PR #378, opened 2016-05-11
Update engagement creation logic.

eschlange:
* Added engagement model.
* Updated navigator controller logic to create engagements upon entry.
* Added specs for the new navigator controller logic.
* Updated task status association with engagement.

[#118962341]


app/controllers/think_feel_do_engine/navigator_controller.rb
```diff
@@ -77,19 +81,54 @@ def show_next_provider
...
       @task_status = membership
                      .task_statuses
-                     .for_content_module(content_module)
-                     .available_by_day(membership.day_in_study)
-                     .try(:order, "start_day DESC")
-                     .try(:first)
+                     .most_recent_for_content_module(content_module,
+                                                     membership.day_in_study)
+                     .first
+      if !@task_status.nil? &&
```
michaelwehrley: Can we instead write `if @task_status && ...`

app/models/engagement.rb
```diff
@@ -0,0 +1,7 @@
+# Used to determine when a patient starts and completes learn task_statuses
+class Engagement < ActiveRecord::Base
+  belongs_to :task_status
+
+  delegate :participant_id, to: :task_status, prefix: false
```
michaelwehrley: I can't find this being used.  Is this legacy?

app/controllers/think_feel_do_engine/navigator_controller.rb
eschlange: Nice catch, I'll update now.

app/controllers/think_feel_do_engine/navigator_controller.rb
```diff
@@ -77,19 +81,54 @@ def show_next_provider
...
+         first_content_slide?(content_module, provider_id, content_id))
+        @task_status
+          .engagements
+          .build
+      end
       @task_status.mark_complete unless @task_status.nil?
     end
 
+    def first_content_slide?(content_module, provider_id, content_id)
+      first_content_provider = content_module
```
michaelwehrley: Would it make sense instead to have a method `first?` on the content_provider itself.

app/controllers/think_feel_do_engine/navigator_controller.rb
```diff
@@ -77,19 +81,54 @@ def show_next_provider
...
+        content_id == FIRST_CONTENT_POSITION
+    end
+
+    def mark_engagement_completed
+      if navigator_content_module && last_engagment
+        last_engagment.update_attributes(completed_at: DateTime.current)
+      end
+    end
+
+    def content_not_specified?(provider_id, content_id)
```
michaelwehrley: I would also recommend changing this to `content_specified?`

```ruby
def content_specified?(provider_id, content_id)
  provider_id && content_id
end

File: app/controllers/think_feel_do_engine/navigator_controller.rb [hunk 5]
```diff
@@ -77,19 +81,54 @@ def show_next_provider
...
+                                                     membership.day_in_study)
+                     .first
+      if @task_status &&
+         (content_not_specified?(provider_id, content_id) ||
+         first_content_slide?(content_module, provider_id, content_id))
+        @task_status
+          .engagements
+          .build
+      end
       @task_status.mark_complete unless @task_status.nil?
```
michaelwehrley: I know this isn't your update.  But it may be more clear if we wrote `@task_status.mark_complete if @task_status`

michaelwehrley:
Things look good.  This is my one real consideration: There are a lot of private methods that are not being tested.  I would move them into their own concern, make them public and test them.  You won't need to put anything in any of the dependent gems, but things would be better tested and easier to read.  👍

eschlange:
Suggested changes were made. I also created a TFD maintenance project story to continue with more complex refactoring: https://www.pivotaltracker.com/story/show/119400247

PR #382, opened 2016-05-23
Ems 118948737

eschlange:
Reduce participant scope for group dash and reports.
* Add scopes to group and participant models.
* Add specs for new scopes.
* Update reports to use the new scope.
* Update views from dashboard to utilize the new scope.
* Update view spec to expect scope usage.
* Update fixtures to include moderator participant and membership.

[#118948737]


app/models/group.rb
```diff
@@ -128,4 +128,12 @@ def likes_by_week(week_number)
                  .lt(membership.week_end_day(week_number)))
     end.map(&:count).sum
   end
+
+  def non_moderator_memberships
+    memberships.joins(:participant)
```
ericcf: What do you think about adding a scope to `Participant` so you can make this refactor:
```ruby
memberships.joins(:participant).merge(Participant.non_moderator)
```
```ruby
class Participant
  scope :non_moderator, lambda { where(is_admin: false) }
end
```
?

app/models/participant.rb
```diff
@@ -92,6 +92,10 @@ def self.inactive
       )
   }
 
+  scope :not_moderator, lambda {
+    where(arel_table[:is_admin].eq(false))
```
ericcf: Could just be `where(is_admin: false)`

spec/models/group_spec.rb
```diff
@@ -136,4 +136,13 @@ def group(attributes = {})
       end
     end
   end
+
+  describe "#non_moderator_memberships" do
+    let(:group) { groups(:group1) }
+
+    it "omits memberships which belong to moderator participants" do
+      expect(group.memberships.count)
+        .to eq(group.non_moderator_memberships.count + 1)
```
ericcf: This is going to be fragile. I'd suggest using the `expect do ... end.to change {}` pattern instead so that it's not dependent on fixtures.

app/models/group.rb
ericcf: I see that you added a scope. You can still use it in the `merge` refactor I suggested.

eschlange:
@ericcf Thanks for the comments. I've made changes applicable to your comments, please take a look when you have a chance.

spec/models/group_spec.rb
```diff
@@ -136,4 +136,15 @@ def group(attributes = {})
       end
     end
   end
+
+  describe "#non_moderator_memberships" do
+    let(:group) { groups(:group1) }
+
+    it "omits memberships which belong to moderator participants" do
```
ericcf: How can you tell that the participant you've created here is a moderator?

spec/models/group_spec.rb
```diff
@@ -136,4 +136,15 @@ def group(attributes = {})
...
+      expect do
+        participant = Participant.create
+        Membership.create(participant: participant, group: group)
+      end.to change { group.non_moderator_memberships.count }.by(0)
```
ericcf: How do you know that this scope ever returns records?

spec/models/group_spec.rb
eschlange: I made updates to ensure creation.

spec/models/group_spec.rb
```diff
@@ -136,4 +136,25 @@ def group(attributes = {})
...
+          email: "big_mods_99@ex.co",
+          password: "oi sRS 89/",
+          password_confirmation: "oi sRS 89/",
+          study_id: "BigMod1",
+          is_admin: true)
+        Membership.create!(participant: participant,
+                           group: group,
+                           start_date: Date.today.advance(days: -2),
+                           end_date: Date.today.advance(days: 10))
+      end.to change { group.non_moderator_memberships.count }.by(0)
```
ericcf: How do you know that `non_moderator_memberships` ever returns associations?

eschlange:
I removed the spec, I usually don't test scopes anyway. I squashed the commits as well. Thanks for looking this over!

ericcf:
💇

PR #384, opened 2016-05-24
Ems 119216231 - Limit emotion name length.

eschlange:
- Add name length validation to emotion model. 
- Limit emotion name length in emotion form. 
[#119216231]


app/models/emotion.rb
```diff
@@ -6,6 +6,7 @@ class Emotion < ActiveRecord::Base
   validates :creator, presence: true
   validates :name, presence: true
   validates :name, uniqueness: { scope: :creator_id }
+  validates :name, length: { maximum: 255 }
```
michaelwehrley: In the future you can chain all of these validations together:
```ruby
validates :name,
          presence: true,
          uniqueness: { scope: :creator_id },
          length: { maximum: 255 }
```

michaelwehrley:
👀  looks good to me 💘

PR #385, opened 2016-05-25
Arm Duplication Method

michaelwehrley:
* Update original copier module
* Update specs
* Update methods that fixed bugs

[#115082907]


michaelwehrley:
@ericcf I worked through a few bugs in this code today while updating it.  Please take a look when you get a chance.  In addition, I already tested it with production data and verified all is good.  Thanks!

app/models/concerns/copier.rb
```diff
@@ -7,6 +7,22 @@ module Copier
     TOOL_TITLE = "LEARN"
```
ericcf: Why is this in the `Concerns` module if it doesn't inherit from the `Concern` class?

app/models/concerns/copier.rb
```diff
@@ -31,6 +47,16 @@ module ::BitCore
       Tool.class_eval do
         LESSON_MODULE = "ContentModules::LessonModule"
```
ericcf: Why not use `ContentModules::LessonModule.to_s` here?

app/models/concerns/copier.rb
```diff
@@ -84,35 +110,38 @@ def copy_source_content_to!(new_arm:, new_content_module:)
...
         def copy_to(new_arm)
           Slideshow.transaction do
-            slideshow = Slideshow.create!(arm: new_arm, title: title)
+            slideshow = Slideshow
+                        .find_or_create_by!(arm_id: new_arm.id, title: title)
 
-            copy_slides_to!(slideshow)
+            copy_slides_to(slideshow)
 
             slideshow
```
ericcf: Are you using this return value?

app/models/concerns/copier.rb
```diff
@@ -84,35 +110,38 @@ def copy_source_content_to!(new_arm:, new_content_module:)
...
             is_title_visible: is_title_visible,
-            options: options,
             position: position,
-            slideshow: slideshow,
+            bit_core_slideshow_id: slideshow.id,
             title: title,
             type: type)
+
+          slide.update_attributes!(options: options)
+          slide
```
ericcf: Are you using this return value?

spec/models/concerns/copier_spec.rb
```diff
@@ -4,6 +4,30 @@ module Concerns
...
+        old_arm.copy_all_to(new_arm)
+      end
+
+      it "copies all tools" do
+        expect(new_arm.bit_core_tools.count)
+          .to eq old_arm.bit_core_tools.count
+      end
+
+      it "copies all content modules" do
+        expect(new_arm.bit_core_tools.map { |tool| tool.content_modules.count }.sum)
```
ericcf: FYI: you don't need to call `.map {}.sum`, you can just call `.sum {}`: http://api.rubyonrails.org/classes/Enumerable.html#method-i-sum

app/models/concerns/copier.rb
michaelwehrley: I was under the impression that we were placing modules in the concerns namespace even if we don't use `extend ActiveSupport::Concern`.  Is that not the case?  Are you saying instead for we just place them in the lib directory etc?

app/models/concerns/copier.rb
michaelwehrley: Is the benefit that there is then a check on whether ContentModules::LessonModule is actually a class or is this just more idiomatic ruby?  Considering I call `# frozen_string_literal: true` above, I would think that from a constant standpoint, it would be better because it isn't malleable.

app/models/concerns/copier.rb
michaelwehrley: I need the slideshow as the return value.

app/models/concerns/copier.rb
michaelwehrley: good catch.

spec/models/concerns/copier_spec.rb
michaelwehrley: that is great!  thanks!

michaelwehrley:
@ericcf please take a look when you get a chance.  Thanks!

ericcf:
:shipit:

PR #386, opened 2016-05-25
Fix engagement completion issue in navigator controller.

eschlange:
* Previously marking engagement complete on any next page.
* Updated logic for checking provider via string comparison.
* Added specs to ensure engagement marked as complete on last page only.

[#120212219]


eschlange:
@michaelwehrley I found an issue with engagement completion logic relating to my recent task status updates. Please let me know what you think.

michaelwehrley:
:nice 🏇

PR #387, opened 2016-05-27
Ecf arel refactor

ericcf:
On a set of test data, this reduces SQL query count from ~5600 to ~2000 and the latency (in development) from ~26s to ~10s. There is room for more improvement, especially in the social networking and lessons tables.


ericcf:
@michaelwehrley @eschlange it would be great if both of you could review this when you have a chance. I made a fair amount of updates and it's possible I missed some things. Thanks!

app/helpers/think_feel_do_engine/coach/group_dashboard_helper.rb
```diff
@@ -145,10 +145,10 @@ def goal_like_count(goal)
       private
 
       def week_of_task(task)
-        if (task.task_statuses.first.start_day / 7.0).ceil == 0
+        if (task.release_day / 7.0).ceil == 0
```
michaelwehrley: Nice; I am not sure why we use `start_day` for task_statuses.

app/models/arm.rb
```diff
@@ -38,7 +38,7 @@ def woz?
   end
 
   def non_home_tools
-    tools = Arel::Table.new(:bit_core_tools)
+    tools = BitCore::Tool.arel_table
```
michaelwehrley: in rails 5 it seems we can look forward to using active record for these types of queries: `BitCore::Tool.where.not(type: "Tools::Home").or(BitCore::Tool.where(type: nil))`

app/models/group_metrics/weekly_count.rb
```diff
@@ -0,0 +1,35 @@
...
+             primary_key: :participant_id
+
+    def self.fetch(group_id)
+      pg_timezone = ActiveSupport::TimeZone[Time.zone.name].tzinfo.name
+
+      joins(memberships: :participant)
+        .select(<<-SQL
+          ( FLOOR ( EXTRACT ( EPOCH FROM (
+            ( #{connection.quote_table_name(table_name)}.created_at
+              AT TIME ZONE 'UTC' ) AT TIME ZONE #{connection.quote(pg_timezone)}
```
michaelwehrley: you added in the time zone since last time we talked.  looks good!

app/models/participant_metrics/weekly_count.rb
```diff
@@ -0,0 +1,38 @@
...
+      joins(:memberships)
+        .select(<<-SQL
+          ( FLOOR ( EXTRACT ( EPOCH FROM (
+            ( #{connection.quote_table_name(table_name)}.created_at
+              AT TIME ZONE 'UTC' ) AT TIME ZONE #{connection.quote(pg_timezone)}
+            - memberships.start_date ) ) / 604800 ) + 1
+          )::int AS week, memberships.participant_id, COUNT(1)
+        SQL
+        )
+        .merge(Membership.where(group_id: group_id))
```
michaelwehrley: Above in `GroupMetrics::WeeklyCount` we excluded the membership that was the moderator.  Is that something we want to do here?

app/views/think_feel_do_engine/coach/group_dashboard/_lessons.html.erb
```diff
@@ -4,10 +4,11 @@
   </div>
 
   <div class="panel-body">
+    <% tasks = group.learning_tasks %>
     <% (1..study_length_in_weeks).each do |week_number| %>
       <h2>Week <%= week_number %></h2>
       <table class="table">
-        <% group.learning_tasks.each do |task| %>
+        <% tasks.each do |task| %>
```
michaelwehrley: nice catch.

michaelwehrley:
These updates and specs look great.  Thanks for the hard work. 👍  🎋

app/helpers/think_feel_do_engine/coach/group_dashboard_helper.rb
ericcf: Possibly for recurring tasks?

app/models/participant_metrics/weekly_count.rb
ericcf: I'm not sure. The participant metrics are broken out by participant, not lumped together like the group metrics so maybe it's not necessary? @eschlange do you have more info here?

eschlange: The recent story/task that I worked on regarding removal of moderators from participant dashboard data was to: "Remove moderator from all counts on dashboard reports". Should I follow up with Marya regarding the use of this WeeklyCount in the admin dashboard pages?

ericcf: Nope, that seems pretty clear. I'll add the extra scope. Thanks.

PR #388, opened 2016-05-27
serialize 'payload' as Hash instead of Parameters

ericcf:


michaelwehrley:
👍  Looks great to me!

eschlange:
Nice work.

PR #389, opened 2016-05-31
Ecf 120467855 site message invalid

ericcf:


app/controllers/think_feel_do_engine/coach/site_messages_controller.rb
```diff
@@ -32,6 +32,7 @@ def create
           redirect_to coach_group_site_message_path(@group, @site_message),
                       notice: "Site message was successfully created."
         else
+          set_participants
```
eschlange: Is this a redundant call? It looks like this may be unnecessary.

eschlange:
👍

michaelwehrley:
👍 looks good to me!

PR #391, opened 2016-05-31
Fix mood and emotion graph visualization bugs.

eschlange:
* Updated the javascript and mood/emotion data collection logic to solve viz bug.
* Moved mood and emotion visualization logic out of participant and into a new service.
* Updated references of visualization logic in existing controller and view logic.

[#119306809]


app/support/mood_and_emotion_visualization_service.rb
```diff
@@ -0,0 +1,66 @@
...
+        }
+        averaged_ratings << daily_negative
+      end
+    end
+    averaged_ratings
+  end
+
+  private
+
+  def average_rating(array)
```
ericcf: Won't this raise a `ZeroDivisionError` if it's an empty array?

eschlange: Yes, but that method never runs if the array has a size of 0.

app/assets/javascripts/think_feel_do_engine/feel/draw_graphs.js
```diff
@@ -25,6 +25,9 @@
         .call(columnChart(graphParameters.startDate, graphParameters.endDate, Y_MIN, Y_MAX, GRAPH_TITLE, YLABEL)
           .width(graphParameters.graphWidth)
           .height(PIXEL_HEIGHT)
+          .x(function(date, i) {
+            return moment(date.day, 'YYYY-MM-DD').startOf('day')._d;
```
ericcf: Why are you accessing `_d` here? It seems risky to depend on a "private" property.

eschlange: Mike, do you know the original purpose of accessing ```_d``` in this code?

michaelwehrley: makes sense; let's not use a private property.  Good catch ⚾️

eschlange: I found the Moment library method specifically designed to return the wrapped javascript Date object. I updated the code to use the method.

ericcf:
I think the remaining item is testing. Are these graphs covered somehow in acceptance testing @cbrenner04 ?

michaelwehrley:
Thanks for moving the code out of the participant class and creating a service.  My only concern would be the lack of specs for this service.  It seems however through conversation with @eschlange that this may be too tedious and unnecessary given we may have additional feature specs via @cbrenner04.

michaelwehrley:
👍

ericcf:
If @cbrenner04 feels that the acceptance tests give us adequate confidence then I agree tests might be unnecessary. I'm not sure what tediousness has to do with it.

cbrenner04:
Sorry for taking so long to respond. I clearly need to update the acceptance tests as this slipped through them before but otherwise I think I can cover it.

PR #392, opened 2016-06-01
consolidate all routing errors and return 404s

ericcf:
[#120467987] [#119117857] [#115254909]


app/controllers/think_feel_do_engine/application_controller.rb
```diff
@@ -13,13 +13,15 @@ class ApplicationController < ::ApplicationController
     INACTIVE_MESSAGE = "We're sorry, but you can't sign in yet "\
               "because you are not assigned to an active group."
 
-    protect_from_forgery with: :exception
+    protect_from_forgery with: :exception, except: :raise_not_found!
```
eschlange: I wasn't familiar with the :except param, its very convenient for this circumstance!

eschlange:
With regards to this change and in terms of the flow of the request, let me know if this is correct:
- a request comes in for an unknown url or action.
- the middle ware throws a "ActionController::RoutingError" exception.
- the ApplicationController recognizes the error before it attempts its "protect_from_forgery" and handles it accordingly

What exactly does ":except" parameter add in the line ```protect_from_forgery with: :exception, except: :raise_not_found!``` if the RoutingError is handled in the before action?

ericcf:
Yes, that description is accurate. Rails won't let us just rescue from the `RoutingError` directly so we need this round about strategy for capturing it.

I added the `:except` because e.g. `PUT /` will trigger a forgery exception in the controller (the CSRF token would be missing) otherwise.

eschlange:
Thanks for the explanation!

PR #393, opened 2016-06-01
Added check for complete task statuses on task delete.

eschlange:
- Updated tasks controller logic to check for complete task statuses.
- Added specs for deletion logic.

[#115860021]


spec/fixtures/tasks.yml
```diff
@@ -10,6 +10,12 @@ task1day2:
   bit_core_content_module: do_awareness
   release_day: 2
 
+task_no_complete:
```
michaelwehrley: @eschlange I don't think you are using this.  Are you?

michaelwehrley:
Everything else looks good 👍

spec/fixtures/tasks.yml
eschlange: @michaelwehrley It is being loaded into the test database and it is being accessed in the modified spec via the query on the Task model.

PR #394, opened 2016-06-03
Fix logout redirect cookie overflow exceptions.

eschlange:
* Override store_location_for Devise function.
* Now doesn't store redirect URL if it causes an exception.
* Added spec that recreates the problem.
* Limit site message message form field.

[#118749637]


ericcf:
Nice! Thanks.

PR #395, opened 2016-06-06
Fix input syntax for maxlength of emotion name.

eschlange:
[#119216231]


eschlange:
Merging, small template change.

PR #396, opened 2016-06-06
Add maxlength to participant message form input field.

eschlange:
[#118749637]


eschlange:
Merging, small template change.

PR #398, opened 2016-06-07
Adding size constraint to coach message body.

eschlange:
[#118749637]


eschlange:
Merging, small template update.

PR #399, opened 2016-06-07
Update subject for participant message emails.

eschlange:
[#120243643]


eschlange:
Similar to previous subject change, merging.

PR #406, opened 2016-06-09
Update default database cleaner strategy. 

eschlange:
- Update default strategy from truncation to transaction.
- Added unique strategy for feature tests.
- Update specs to accommodate database cleaner config changes.
- Update participant model to handle missing fields on participant.
- Update specs to remove models between specs.
- Updated participant method and assocaited spec to handle missing events.
- Removed unnecessary patient dash spec (scope already tested in activity spec line 109).
- Wiping table for feature test.
- Destroy existing moods to eliminate intermittent spec errors.

[#118847783]


spec/rails_helper.rb
```diff
@@ -59,7 +59,7 @@
   end
 
   config.before(:each) do
-    DatabaseCleaner.strategy = :truncation
+    DatabaseCleaner.strategy = :transaction
```
michaelwehrley: If we are using only transactions now, can we remove `DatabaseCleaner` all together?

spec/models/thought_spec.rb
```diff
@@ -8,6 +8,10 @@
                         email: "e@foo.bar")
   end
 
+  before :each do
+    Thought.destroy_all
```
michaelwehrley: I was wondering if it is reasonable to update the specs so we don't have to use `destroy_all` before them each.  Is that possible or too much overhead?

spec/rails_helper.rb
eschlange: When I remove the configuration to stop using truncation for the :js specs, we get a fairly large number of failures. If we remove the configuration, I'm not sure how we'd update that specific strategy.

spec/models/thought_spec.rb
eschlange: Any recommendations on how to do so? I'm not sure how that can be accomplished.

michaelwehrley: We could make that for another story I guess.  We would have to go through each failing spec in the model spec.

spec/rails_helper.rb
michaelwehrley: Okay.  Great.  Thanks!

michaelwehrley:
Nice updates.  Thanks for looking at this!

michaelwehrley:
👍

spec/models/thought_spec.rb
eschlange: I'll add a TFD maintenance story to look into this.

PR #408, opened 2016-06-10
experiment with fixutes for some specs

ericcf:
This uses transactional fixtures with the non-feature specs, and database cleaner with just the feature specs. It decreases the Travis build time by 40-50%.


spec/requests/application_requests_spec.rb
```diff
@@ -1,4 +1,4 @@
-require "rails_helper"
+require "rails_fixture_helper"
```
michaelwehrley: Thanks for sharing these `request` specs.

spec/rails_fixture_helper.rb
```diff
@@ -0,0 +1,60 @@
...
+  config.after :suite do
+    Timecop.return
+  end
+
+  # If true, the base class of anonymous controllers will be inferred
+  # automatically. This will be the default behavior in future versions of
+  # rspec-rails.
+  config.infer_base_class_for_anonymous_controllers = false
+
+  config.include AuthenticationHelpers, type: :feature
```
michaelwehrley: Do you need these in here if this is for fixtures?

eschlange:
This looks great, I give the OK to merge. I'll do some local test runs as well and compare them against my changes.

spec/rails_fixture_helper.rb
ericcf: Nope! I'll remove.

PR #411, opened 2016-06-14
Update activities page logic to stop multiple form submissions.

eschlange:
[#121498511]


eschlange:
Merging, small Javascript change.

PR #413, opened 2016-06-16
convert Slide options to Hashes

ericcf:
* they are currently Parameters, which are not forwards compatible (with
  Rails 5)


eschlange:
I see now~ 👍

PR #414, opened 2016-06-16
convert existing Events to use Hash for payloads

ericcf:
* this is forwards compatible with Rails 5


db/migrate/20160616201120_convert_event_payloads.rb
```diff
@@ -0,0 +1,23 @@
...
+      # Use 'update_column' because with 'update' ActiveRecord will try to
+      # deserialize the existing value first, and will detect that there are
+      # two identical Hashes. Use 'to_unsafe_h' because otherwise all keys
+      # will be deemed unsafe, having not been permitted via the strong
+      # parameters API.
+      event.update_column :payload, event.payload.to_unsafe_h
+    end
+  end
+
+  def down
```
michaelwehrley: I think you can leave `down` out and you will still be able to roll the migraiton back.

michaelwehrley:
👍 looks good to me

PR #415, opened 2016-06-17
fix sporadically failing password spec

ericcf:
* delete the test for medium strength passwords because it sporadically
  fails; we're already testing weak and strong which seems like enough
  for a feature spec
* consolidate multiple specs testing the strong password case
* run blank password spec without JavaScript for efficiency


cbrenner04:
👍

PR #418, opened 2016-06-20
Update activity rows in summary table group dash

cbrenner04:
[Finishes #119923709]


ericcf:
I'm not at my computer, but I like your improved solution! Looks good to merge.

cbrenner04:
@ericcf thanks. I'll merge it. The counts weren't updating with our first solution so I asked @eschlange for some help and this is what we ended up with.

PR #419, opened 2016-06-23
Add Try to Active Memberhip

michaelwehrley:
* Add try method to active membership when asking for a group

[Finishes: #121680805]


a-ashbeck:
Looks good.

PR #420, opened 2016-06-24
Update When Engagements are Updated

michaelwehrley:
* Engagement updated only when the last slideshow provider of the engagement is viewed
* Add conditional specs for last slideshow provider of an 'engagement'

[#120722739]


eschlange:
Looks solid, nice tests.

PR #421, opened 2016-07-12
Catch Twilio Errors

michaelwehrley:
* Catch and save errors for reporting
* Add model spec
* Update MessageSmsNotification spec
* Add reporting functionality
* Add report spec

[Finishes: #121225259]


ericcf:
I'm curious - why create a report for this rather than using Sentry?

michaelwehrley:
I thought it would be a good way for users to see who has unsubscribed.

ericcf:
But this doesn't discriminate between subscription exceptions and other exceptions does it? It seems bad to swallow all exceptions. We would never know there was a problem unless we manually checked the exports right?

michaelwehrley:
You make a good point that it would [hide some other 400 and 500](https://github.com/twilio/twilio-ruby/blob/master/docs/usage/errors.rst) errors that we want to be notified about.  However, do we really want to be notified every day for 7 weeks while someone is unsubscribed?  Would a better solution be to have both?  Record it and send a Sentry message?

ericcf:
Can you explain why an exception is occurring? I still don't understand that part. Thanks.

PR #423, opened 2016-07-14
ensure activity submit is disabled unless it fails

ericcf:
[#121498511]


michaelwehrley:
Seems reasonable.  Nice work.

PR #424, opened 2016-07-14
ensure headers are deserialized properly

ericcf:
* configure and test strong parameters

[#124931341]


michaelwehrley:
looks good to me! 👍

PR #428, opened 2016-07-18
augment report data

ericcf:
[#122733031]


michaelwehrley:
Looks like good updates 👍  Thanks!

PR #429, opened 2016-07-22
Ems 124235441 - Updated scope for planned activities in participant view.

eschlange:
- Added model scope.
- Updated content provider view.
- Added fixture to test new scope.
- Added spec for new scope.

[#124235441]


app/models/content_providers/unplanned_activity_form.rb
```diff
@@ -12,7 +12,7 @@ def render_current(options)
                                   .view_context
                                   .current_participant
                                   .activities
-                                  .in_the_future
+                                  .starts_in_the_future
```
michaelwehrley: So we are no longer showing activities that end in the future, but have already started?

eschlange: That is correct, see Chris's response here: https://www.pivotaltracker.com/story/show/124235441

michaelwehrley: Awesome!  Thanks.

michaelwehrley:
👍

PR #433, opened 2016-12-17
Add Validation for Title

michaelwehrley:
* Add Activity Type validation for title

[#132164171]


spec/models/activity_type_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+  fixtures :all
+
+  describe "validations" do
+    let(:participant) { participants(:participant1) }
+
+    it "validates the length of the title" do
+      expect(
+        participant
+          .activity_types
+          .build(title: "Foo" * 100)
```
ericcf: What is the purpose of using fixtures? Why not just

```ruby
ActivityType.new(title: "X" * 256).tap(&:valid?).errors.full_messages
```

